PR #59, opened 2017-03-29
Extend SCIM2.0 metadata to include User Lifecycle State

IndunilRathnayake:


IndunilRathnayake:
Some of the SCIM test cases are failing due to identity.mgt version upgrade, need to check that:  https://wso2.org/jira/browse/IDENTITY-5817

PR #80, opened 2017-10-04
IDENTITY-6611: Add initial unit test cases for scim2 common utils

ashensw:


pulasthi7:
Merging since this is a blocker to others. Will address the issues of these PR in a separate one.

PR #97, opened 2018-03-05
Hide service listing and upgrade cxf version

GDRDABARERA:
Hide service listing in SCIM2 and upgrade cxf version


GDRDABARERA:
The travis build has failed, at that time the new orbit bundle was not released. But now it has released and the clean repo build passes

PR #120, opened 2018-10-23
Fix users can be deleted by the SCIM 2 'Delete User' service without enabling the SCIM

isharailanga:
This PR is to fix https://github.com/wso2/product-is/issues/3748

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -252,6 +252,12 @@ public class SCIMUserManager implements UserManager {
             } else {
                 //we assume (since id is unique per user) only one user exists for a given id
                 userName = userNames[0];
+                //check if SCIM is enabled for the user store
+                String userStoreDomainName = IdentityUtil.extractDomainFromName(userName);
+                if (StringUtils.isNotBlank(userStoreDomainName) && !isSCIMEnabled(userStoreDomainName)) {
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -252,6 +252,12 @@ public void deleteUser(String userId) throws NotFoundException, CharonException
             } else {
                 //we assume (since id is unique per user) only one user exists for a given id
                 userName = userNames[0];
+                //check if SCIM is enabled for the user store
```
DMHP: Move the comment to a line below, just before the if condition.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -252,6 +252,12 @@ public void deleteUser(String userId) throws NotFoundException, CharonException
...
+                String userStoreDomainName = IdentityUtil.extractDomainFromName(userName);
+                if (StringUtils.isNotBlank(userStoreDomainName) && !isSCIMEnabled(userStoreDomainName)) {
```
IsuraD: Is there any possibility of userStoreDomainName becomes empty?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
isharailanga: 3b71977a5ec04cda298019f05419d4b051dcd457

PR #129, opened 2019-02-08
Fix  wso2/product-is#3847

IsurangaPerera:
Fix wso2/product-is#3847

This problem was caused by 2 issues in the implementation

1. When contains operator is used it creates a modified value by closing the value with wild card characters. As an example, SECONDARY/username will be transformed to *SECONDARY/username*. PR solves this issue by correcting it as SECONDARY/*username*. 

2. When using JDBC user stores along with SCIM implementation will skip username checks. This issue has been addressed by https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/132


malithie:
@IsurangaPerera ,
Can we please fill up the approach you take on fixing this respective issue and what caused it in high level

malithie:
@IsurangaPerera , Can we incorporate a unit test for this case.

PR #132, opened 2019-02-12
Fix SCIM2 lists admin in full list, but not when specifically requested

emswbandara:
Fixes wso2/product-is#4415


tharindu-bandara:
Fixes https://github.com/wso2/product-is/issues/6025

PR #149, opened 2019-04-11
Introducing a new function to be used for the filtering scenario and listing users scenario.

somindatommy:
Both the filtering scenario and listing users scenario uses the same logic to retrieve users. Therefore a common method is introduced.

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -784,43 +784,68 @@ public class SCIMUserManager implements UserManager {
             return getFilteredUsernames(createConditionForSingleAttributeFilter(domainName, node), offset, limit,
                     sortBy, sortOrder, domainName);
         } else {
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -784,43 +784,68 @@ private boolean isDomainNameEmbeddedInAttributeValue(String filterOperation, Str
...
+    private String[] filterUsersFromMultipleDomains(Node node, int offset, int limit, String sortBy, String sortOrder,
+            Condition conditionForListingUsers) throws CharonException {
+
+        // Filter users when the domain is not set in the request. Then filter through multiple domains.
+        String[] userStoreDomainNames = getDomainNames();
+        ArrayList<String> filteredUsernames = new ArrayList<>();
+        Condition condition;
+        for (String userStoreDomainName : userStoreDomainNames) {
+
+            // Check whether the used case is for listing users.
```
vihanga-liyanage: Type: Used case

somindatommy: https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/218

PR #158, opened 2019-04-24
Fix #5088:SCIM2 user listing with an offset for a secondary user store is not working properly

IsurangaPerera:
- Fix wso2/product-is#5088
The issue was caused by non-scim users user being added to the result set. Since they don't have a scim id it returns "invalidValue","detail":"Required attribute id is missing in the SCIM Object." response.

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -360,8 +360,9 @@ public class SCIMUserManager implements UserManager {
                 log.debug(message);
             }
         } else {
-            users.set(0, userNames.length); // Set total number of results to 0th index.
-            users.addAll(getUserDetails(userNames, requiredAttributes)); // Set user details from index 1.
+            List<Object> scimUsers = getUserDetails(userNames, requiredAttributes);
+            users.set(0, scimUsers.size()); // Set total number of results to 0th index.
+            users.addAll(scimUsers); // Set user details from index 1.
         }
         return users;
     }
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -538,7 +539,7 @@ private void retriveSCIMUsers(List<Object> users, String[] userNames, List<Strin
                             + "Including user : " + userName + " in the response.");
                 }
                 User scimUser = this.getSCIMUser(userName, requiredClaims, scimToLocalClaimsMap);
-                if (scimUser != null) {
+                if (scimUser != null && StringUtils.isNotBlank(scimUser.getId())) {
```
malithie: Shouldn't this be handled within the getSCIMUser it self similar to [1].
That's seems to be the core method being used to build up the user from SCIM schema and being referred in other places. Fixing only this place will not fix any other places where the same issue could occur

[1] https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/blob/bb3cbb2d11fa4e2940c3fec29f662a0c21f1c21f/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java#L2157

IsurangaPerera: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2055,6 +2056,10 @@ private User getSCIMUser(String userName, List<String> claimURIList, Map<String,
             Map<String, String> attributes = SCIMCommonUtils.convertLocalToSCIMDialect(userClaimValues,
                     scimToLocalClaimsMap);
 
+            if (!attributes.containsKey(SCIMConstants.CommonSchemaConstants.ID_URI)) {
+                return scimUser;
+            }
```
malithie: With this fix there's a possibility to return null from this method.
Make sure invocations of this method handles that

IsurangaPerera: it's already handled

PR #160, opened 2019-04-25
Inconsistency in SCIM2 Filtering with domain parameter and domain in the attribute value

somindatommy:
Resolve: [wso2_product_is#5129](https://github.com/wso2/product-is/issues/5129)

The original code supported embedding domain name only in the username and groups. The fix changes checking for a group uri or a username uri by evaluating all the search attributes.


somindatommy:
It was decided to only let userName and group attributes to embed domain name in the attribute value. Therefore the above fix is not needed.

PR #161, opened 2019-04-25
Fix SCIMCommonUtils.isFilteringEnhancementsEnabled() property is used in both request flow and response flow for two different behaviours

IshikaGodage:
Fix [wso2/product-is #4961](https://github.com/wso2/product-is/issues/4961)

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -705,8 +705,6 @@ public class SCIMUserManager implements UserManager {
                     node.getValue()));
         }
 
-        // Apply filter enhancements for username and update expression node value.
-        applyFilterEnhancementsForUsernameInNode(node);
         try {
             // Extract he domain name if the domain name is embedded in the filter attribute value.
             domainName = resolveDomainNameInAttributeValue(domainName, node);
@@ -715,6 +713,10 @@ public class SCIMUserManager implements UserManager {
                     .format("Domain parameter: %s in request does not match with the domain name in the attribute "
                             + "value: %s ", domainName, node.getValue());
             throw new CharonException(errorMessage, e);
+        }
+        // Get domain name according to Filter Enhancements properties as in identity.xml
+        if (StringUtils.isEmpty(domainName)) {
+            domainName = getDomainNameAccordingToFilterEnhancementProperty(node);
         }
         try {
             // Check which APIs should the filter needs to follow.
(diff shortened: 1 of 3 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -731,6 +726,32 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
...
+     *
+     * @param domainName Domain to run the filter
+     * @param node       Expression node for single attribute filtering
+     * @return Resolved domainName
+     * @throws CharonException
+     */
+    private String getDomainName(String domainName, ExpressionNode node) throws CharonException {
+
+        try {
+            // Extract he domain name if the domain name is embedded in the filter attribute value.
```
Abilashini: "he"? Please Correct the comment

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2276,13 +2310,31 @@ private Group getGroupWithName(String groupName)
...
+            if (StringUtils.isNotEmpty(userStoreDomainName) && (userStoreDomainName.toUpperCase()
+                    .equals(UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME)) && (
+                    groupName.indexOf(CarbonConstants.DOMAIN_SEPARATOR) > 0)) {
+                int indexOfDomainSeparator = groupName.indexOf(CarbonConstants.DOMAIN_SEPARATOR);
+                String role = groupName.substring(indexOfDomainSeparator + 1);
+                group.setDisplayName(role);
+            } else {
+                group.setDisplayName(groupName);
+            }
+        }
```
malithie: From this block seems even for primary user store user store domain is prepended and it's being removed again here.
From the user core level usually primary is not appended for a domain. That means from the SCIM level itself is should have been added.
Rather being removing an added one better to avoid prepending the domain itself

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1763,7 +1791,9 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
             throw new NotImplementedException(error);
         }
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled()) {
+        if (isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) {
+            attributeValue = prependPrimaryDomain(attributeValue);
```
malithie: For groups domains should be prepended only for group display name right
Won't this impact any other attribute

PR #163, opened 2019-04-26
FIX user specifying count zero

somindatommy:
- FIX [wso2_product_is#5106](https://github.com/wso2/product-is/issues/5106)
- Previously int variables were used to hold values for count and startIndex. Therefore, variables could not hold Null state values.
 - In the fix Count and startIndex variables were set as Integer objects so that they can hold Null state values.
- A new method was declared to process the count and startIndex params in the groups endpoint.

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -310,6 +310,24 @@ public class SCIMUserManager implements UserManager {
     }
 
     @Override
+    public List<Object> listUsersWithGET(Node rootNode, Integer startIndex, Integer count, String sortBy,
+            String sortOrder, String domainName, Map<String, Boolean> requiredAttributes)
+            throws CharonException, NotImplementedException {
+
+        // Validate NULL value for startIndex.
+        startIndex = handleStartIndexEqualsNULL(startIndex);
+        if (sortBy != null || sortOrder != null) {
+            throw new NotImplementedException("Sorting is not supported");
+        } else if (count != null && count == 0) {
+            return Collections.emptyList();
+        } else if (rootNode != null) {
+            return filterUsers(rootNode, requiredAttributes, startIndex, count, sortBy, sortOrder, domainName);
+        } else {
+            return listUsers(requiredAttributes, startIndex, count, sortBy, sortOrder, domainName);
+        }
+    }
+
+    @Override
     public List<Object> listUsersWithPost(SearchRequest searchRequest, Map<String, Boolean> requiredAttributes)
             throws CharonException, NotImplementedException, BadRequestException {
         return listUsersWithGET(searchRequest.getFilter(), searchRequest.getStartIndex(), searchRequest.getCount(),
@@ -329,12 +347,15 @@ public class SCIMUserManager implements UserManager {
      * @return User list with detailed attributes
      * @throws CharonException Error while listing users
      */
-    private List<Object> listUsers(Map<String, Boolean> requiredAttributes, int offset, int limit, String sortBy,
-            String sortOrder, String domainName) throws CharonException {
+    private List<Object> listUsers(Map<String, Boolean> requiredAttributes, int offset, Integer limit,
+            String sortBy, String sortOrder, String domainName) throws CharonException {
 
         List<Object> users = new ArrayList<>();
-        //0th index is to store total number of results.
+        // 0th index is to store total number of results.
         users.add(0);
+
+        // Handle limit equals NULL scenario.
+        limit = handleLimitEqualsNULL(limit);
         String[] userNames;
         if (StringUtils.isNotEmpty(domainName)) {
             if (canPaginate(offset, limit)) {
@@ -648,21 +669,39 @@ public class SCIMUserManager implements UserManager {
     }
 
     /**
+     * Method to handle limit equals NULL in a request.
+     *
+     * @param limit Limit in the request.
+     * @return Updated limit.
+     */
+    private int handleLimitEqualsNULL(Integer limit) {
+
+        // Limit equal to null implies return all users. Return all users scenario handled by the following methods by
+        // expecting count as zero.
+        if (limit == null) {
+            limit = 0;
+        }
+        return limit;
+    }
+
+    /**
      * Filter users using multi-attribute filters or single attribute filters with pagination.
      *
      * @param node               Node
      * @param requiredAttributes Required attributes
      * @param offset             Starting index of the count
      * @param limit              Number of required results (count)
-     * @param sortBy             Sortby
+     * @param sortBy             SortBy
      * @param sortOrder          Sort order
      * @param domainName         Domain that the filter should perform
      * @return Detailed user list
      * @throws CharonException
      */
-    private List<Object> filterUsers(Node node, Map<String, Boolean> requiredAttributes, int offset, int limit,
+    private List<Object> filterUsers(Node node, Map<String, Boolean> requiredAttributes, int offset, Integer limit,
             String sortBy, String sortOrder, String domainName) throws CharonException {
 
+        // Handle limit equals NULL scenario.
+        limit = handleLimitEqualsNULL(limit);
         // Handle single attribute search.
         if (node instanceof ExpressionNode) {
             return filterUsersBySingleAttribute((ExpressionNode) node, requiredAttributes, offset, limit, sortBy,
@@ -734,7 +773,8 @@ public class SCIMUserManager implements UserManager {
     /**
      * Method to decide whether to use new APIs or legacy APIs.
      *
-     * @param limit Number of results required for the filter request.
+     * @param limit Number of results required for the filter request (limit equals to ZERO will retrieve all users
+     *              who matches the filter).
      * @return True if legacy API filtering is needed.
      */
     private boolean isUseLegacyAPIs(int limit) {
@@ -1161,17 +1201,14 @@ public class SCIMUserManager implements UserManager {
      * @return
      * @throws CharonException
      */
-    private List<Object> getMultiAttributeFilteredUsers(Node node, Map<String, Boolean> requiredAttributes,
-                                                        int offset, int limit, String sortBy, String sortOrder,
-                                                        String domainName, List<Object> filteredUsers)
+    private List<Object> getMultiAttributeFilteredUsers(Node node, Map<String, Boolean> requiredAttributes, int offset,
+            int limit, String sortBy, String sortOrder, String domainName, List<Object> filteredUsers)
             throws CharonException {
 
         String[] userNames;
-
         // Handle pagination.
         if (limit > 0) {
-            userNames = getFilteredUsersFromMultiAttributeFiltering(node, offset, limit, sortBy,
-                    sortOrder, domainName);
+            userNames = getFilteredUsersFromMultiAttributeFiltering(node, offset, limit, sortBy, sortOrder, domainName);
             filteredUsers.set(0, userNames.length);
             filteredUsers.addAll(getFilteredUserDetails(userNames, requiredAttributes));
         } else {
@@ -1698,15 +1735,51 @@ public class SCIMUserManager implements UserManager {
     public List<Object> listGroupsWithGET(Node rootNode, int startIndex, int count, String sortBy, String sortOrder,
             String domainName, Map<String, Boolean> requiredAttributes)
             throws CharonException, NotImplementedException, BadRequestException {
+
         if (sortBy != null || sortOrder != null) {
             throw new NotImplementedException("Sorting is not supported");
-        } else if (startIndex != 1) {
+        } else if (startIndex != 1 && count >= 0) {
             throw new NotImplementedException("Pagination is not supported");
         } else if (rootNode != null) {
             return filterGroups(rootNode, requiredAttributes);
         } else {
             return listGroups(requiredAttributes);
         }
+    }
+
+    @Override
+    public List<Object> listGroupsWithGET(Node rootNode, Integer startIndex, Integer count, String sortBy,
+            String sortOrder, String domainName, Map<String, Boolean> requiredAttributes)
+            throws CharonException, NotImplementedException {
+
+        // Validate NULL value for startIndex.
+        startIndex = handleStartIndexEqualsNULL(startIndex);
+        if (sortBy != null || sortOrder != null) {
+            throw new NotImplementedException("Sorting is not supported");
+        } else if (startIndex != 1 || count != null) {
+            throw new NotImplementedException("Pagination is not supported");
+        } else if (rootNode != null) {
+            return filterGroups(rootNode, requiredAttributes);
+        } else {
+            return listGroups(requiredAttributes);
+        }
+    }
+
+    /**
+     * Method to interpret startIndex as 1 when the startIndex equals to NULL in the request.
+     *
+     * @param startIndex StartIndex in the request.
+     * @return Updated startIndex
+     */
+    private int handleStartIndexEqualsNULL(Integer startIndex) {
+
+        if (startIndex == null) {
+            if (log.isDebugEnabled()) {
+                log.debug("NULL value for startIndex argument interpreted as 1");
+            }
+            return 1;
+        }
+        return startIndex;
     }
 
     private List<Object> listGroups(Map<String, Boolean> requiredAttributes) throws CharonException {
@@ -2387,7 +2460,7 @@ public class SCIMUserManager implements UserManager {
      * Paginate a list of users names according to a given offset and a count.
      *
      * @param users  A list of unpaginated users.
-     * @param limit  The total number of results required (Zero will return all the users).
+     * @param limit  The total number of results required (ZERO will return all the users).
      * @param offset The starting index of the count (limit).
      * @return A list of paginated users
      */
@@ -2404,7 +2477,7 @@ public class SCIMUserManager implements UserManager {
             offset = 1;
         }
 
-        // If the results is less than the offset return an empty user list.
+        // If the results are less than the offset, return an empty user list.
         if (offset > users.length) {
             return new String[0];
         }
--- a/components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
+++ b/components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
@@ -18,9 +18,9 @@
 
 package org.wso2.carbon.identity.scim2.provider.resources;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.wso2.carbon.context.PrivilegedCarbonContext;
 import org.wso2.carbon.identity.jaxrs.designator.PATCH;
 import org.wso2.carbon.identity.scim2.common.impl.IdentitySCIMManager;
 import org.wso2.carbon.identity.scim2.provider.util.SCIMProviderConstants;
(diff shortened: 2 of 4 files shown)


malithie:
Depends on https://github.com/wso2/charon/pull/229

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -374,7 +393,7 @@ public void deleteUser(String userId) throws NotFoundException, CharonException
      * @param limit  Counting value
      * @return true if pagination is possible, false otherwise
      */
-    private boolean canPaginate(int offset, int limit) {
+    private boolean canPaginate(int offset, Integer limit) {
```
malithie: Is it necessary to change this signature

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -444,7 +463,7 @@ private boolean canPaginate(int offset, int limit) {
      * @return Paginated usernames list
      * @throws CharonException Pagination not support
      */
-    private String[] listUsernamesAcrossAllDomains(int offset, int limit, String sortBy, String sortOrder)
+    private String[] listUsernamesAcrossAllDomains(int offset, Integer limit, String sortBy, String sortOrder)
```
malithie: Is it necessary to change this one

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -697,7 +728,7 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
      * @throws CharonException Error while filtering
      */
     private List<Object> filterUsersBySingleAttribute(ExpressionNode node, Map<String, Boolean> requiredAttributes,
-            int offset, int limit, String sortBy, String sortOrder, String domainName) throws CharonException {
+            int offset, Integer limit, String sortBy, String sortOrder, String domainName) throws CharonException {
```
malithie: Is it necessary to change the signature here

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -734,14 +765,15 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
...
+     *              who matches the filter).
      * @return True if legacy API filtering is needed.
      */
-    private boolean isUseLegacyAPIs(int limit) {
+    private boolean isUseLegacyAPIs(Integer limit) {
 
         // If the limit is not specified, list all the users using old APIs since the new APIs must have a
         // limit larger than zero.
-        if (limit <= 0) {
+        if (limit == 0) {
```
malithie: I don't see a necessity to change the signature

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1080,7 +1112,7 @@ private Condition createConditionForSingleAttributeFilter(String domainName, Nod
      * @throws NotImplementedException Not supported filter operation
      * @throws UserStoreException
      */
-    private String[] filterUsersUsingLegacyAPIs(ExpressionNode node, int limit, int offset, String domainName)
+    private String[] filterUsersUsingLegacyAPIs(ExpressionNode node, Integer limit, int offset, String domainName)
```
malithie: Is it necessary to change the signature

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1162,18 +1194,18 @@ private Condition createConditionForSingleAttributeFilter(String domainName, Nod
      * @throws CharonException
      */
     private List<Object> getMultiAttributeFilteredUsers(Node node, Map<String, Boolean> requiredAttributes,
-                                                        int offset, int limit, String sortBy, String sortOrder,
+                                                        int offset, Integer limit, String sortBy, String sortOrder,
```
malithie: Is it necessary to change the signature

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1697,10 +1729,28 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
...
+
+    @Override
+    public List<Object> listGroupsWithGET(Node rootNode, Integer startIndex, Integer count, String sortBy,
+            String sortOrder, String domainName, Map<String, Boolean> requiredAttributes)
+            throws CharonException, NotImplementedException, BadRequestException {
+
+
+        if (sortBy != null || sortOrder != null) {
+            throw new NotImplementedException("Sorting is not supported");
+        } else if (startIndex != 1  || count != null) {
```
malithie: As both startIndex and count are instances of Integer it would be better to check for not null of each

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2387,11 +2437,11 @@ private Group addDomainToUserMembers(Group group, String userStoreDomain) throws
      * Paginate a list of users names according to a given offset and a count.
      *
      * @param users  A list of unpaginated users.
-     * @param limit  The total number of results required (Zero will return all the users).
+     * @param limit  The total number of results required (ZERO will return all the users).
      * @param offset The starting index of the count (limit).
      * @return A list of paginated users
      */
-    private String[] paginateUsers(String[] users, int limit, int offset) {
+    private String[] paginateUsers(String[] users, Integer limit, int offset) {
```
malithie: Is it necessary to change the method signature. I don't find a reason here

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed:8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed:8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 8f8c20292c4fac76778a8aa95490675023d7d063

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -647,22 +666,34 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
...
 
+    private int handleLimitEqualsNULL(Integer limit){
+
+        // Limit equal to null implies return all users. Return all users scenario handled by the following methods by
+        // expecting count as zero.
+        if (limit == null) {
+            limit = 0;
+        }
+        return limit;
+
```
sarubi: shall we remove unnecessary lines

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1697,10 +1726,29 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
...
+    @Override
+    public List<Object> listGroupsWithGET(Node rootNode, Integer startIndex, Integer count, String sortBy,
+            String sortOrder, String domainName, Map<String, Boolean> requiredAttributes)
             throws CharonException, NotImplementedException, BadRequestException {
+
         if (sortBy != null || sortOrder != null) {
             throw new NotImplementedException("Sorting is not supported");
-        } else if (startIndex != 1) {
+        } else if(startIndex == null){
+            throw new BadRequestException("NULL argument for startIndex");
```
sarubi: startIndex is an optional argument so isn't it can't be null? Here null means user didn't specify right?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -647,22 +666,34 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
         }
     }
 
+    private int handleLimitEqualsNULL(Integer limit){
+
+        // Limit equal to null implies return all users. Return all users scenario handled by the following methods by
```
tharindu-bandara: line exceeds max length

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -329,12 +345,15 @@ public void deleteUser(String userId) throws NotFoundException, CharonException
...
+    private List<Object> listUsers(Map<String, Boolean> requiredAttributes, int offset, Integer limit,
+            String sortBy, String sortOrder, String domainName) throws CharonException {
 
         List<Object> users = new ArrayList<>();
-        //0th index is to store total number of results.
+        // 0th index is to store total number of results.
         users.add(0);
+
+        // Handle limit equals NULL scenario.
+        limit = handleLimitEqualsNULL(limit);
```
tharindu-bandara: From where will it handle limit < 0 scenario?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1161,17 +1193,14 @@ private Condition createConditionForSingleAttributeFilter(String domainName, Nod
      * @return
      * @throws CharonException
      */
-    private List<Object> getMultiAttributeFilteredUsers(Node node, Map<String, Boolean> requiredAttributes,
-                                                        int offset, int limit, String sortBy, String sortOrder,
-                                                        String domainName, List<Object> filteredUsers)
+    private List<Object> getMultiAttributeFilteredUsers(Node node, Map<String, Boolean> requiredAttributes, int offset,
```
tharindu-bandara: Exceeds max line length

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1161,17 +1193,14 @@ private Condition createConditionForSingleAttributeFilter(String domainName, Nod
...
+            int limit, String sortBy, String sortOrder, String domainName, List<Object> filteredUsers)
             throws CharonException {
 
         String[] userNames;
-
         // Handle pagination.
         if (limit > 0) {
-            userNames = getFilteredUsersFromMultiAttributeFiltering(node, offset, limit, sortBy,
-                    sortOrder, domainName);
+            userNames = getFilteredUsersFromMultiAttributeFiltering(node, offset, limit, sortBy, sortOrder, domainName);
```
tharindu-bandara: Exceeds line length

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
tharindu-bandara: @sarubi +1. Better to use a debug log for dev related information, and end-user friendly message in the returned exception.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2404,13 +2452,13 @@ private Group addDomainToUserMembers(Group group, String userStoreDomain) throws
             offset = 1;
         }
 
-        // If the results is less than the offset return an empty user list.
+        // If the results are less than the offset or the limit (count) equals to zero, return an empty user list.
         if (offset > users.length) {
```
tharindu-bandara: Does this only check for results are less than the offset? if so the  ` or the limit (count) equals to zero,` mentioned in the comment above is misleading

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2404,13 +2452,13 @@ private Group addDomainToUserMembers(Group group, String userStoreDomain) throws
...
             return new String[0];
         }
 
         // If the limit is zero, all the users needs to be returned after verifying the offset.
-        if (limit <= 0) {
+        if (limit == 0) {
```
tharindu-bandara: Does the negative case is handled prior to this method?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: The formatter does not recognize it has an exceeded. It seems here that it exceeds the limit. But in IDEA it isn't.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: The formatter does not recognize it has an exceeded. It seems here that it exceeds the limit. But in IDEA it isn't.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: It is handled from Charon ResourceManagerUtil.processCount(Integer) method

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: The formatter does not recognize it has an exceeded. It seems here that it exceeds the limit. But in IDEA it isn't.

components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
```diff
@@ -312,59 +312,76 @@ private Response processRequest(final Map<String, String> requestAttributes) {
...
-
             return SupportUtils.buildResponse(scimResponse);
-
         } catch (CharonException e) {
             return handleCharonException(e, encoder);
         }
     }
 
+    /**
+     * Method to process the count and startIndex in the request.
```
tharindu-bandara: Shall we change the doc comment, since the method itself is not specific for `count and startIndex`.

components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
```diff
@@ -312,59 +312,76 @@ private Response processRequest(final Map<String, String> requestAttributes) {
...
-
                 String sortBy = requestAttributes.get(SCIMProviderConstants.SORT_BY);
                 String sortOrder = requestAttributes.get(SCIMProviderConstants.SORT_ORDER);
                 String domainName = requestAttributes.get(SCIMProviderConstants.DOMAIN);
 
-                scimResponse = groupResourceManager.listWithGET(userManager, filter, startIndex,
-                        count, sortBy, sortOrder, domainName, attributes, excludedAttributes);
-
+                // Processing count and startIndex in the request.
+                Integer startIndex = processPaginationParamsInRequest(
```
tharindu-bandara: Does the `processPaginationParamsInRequest` method is itself process only the integer parameters? if so shall we change the method name to reflect that?. And since this method process only a single parameter, better to reflect that as well

somindatommy: Fixed: df77256698c74c829147f73d00e2f63979199c16

components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
somindatommy: fixed: df77256698c74c829147f73d00e2f63979199c16

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: It is handled by the charon endpoint. But validating here also would be good.
fixed: df77256698c74c829147f73d00e2f63979199c16

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: df77256698c74c829147f73d00e2f63979199c16

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: Fixed: df77256698c74c829147f73d00e2f63979199c16

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: Fixed: df77256698c74c829147f73d00e2f63979199c16

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -309,6 +309,24 @@ public void deleteUser(String userId) throws NotFoundException, CharonException
         }
     }
 
+    @Override
+    public List<Object> listUsersWithGET(Node rootNode, Integer startIndex, Integer count, String sortBy,
+            String sortOrder, String domainName, Map<String, Boolean> requiredAttributes)
+            throws CharonException, NotImplementedException {
+
+        // Validate NULL value for startIndex.
+        startIndex = handleStartIndexEqualsNULL(startIndex);
```
sarubi: We already did the null checks in Charon under processStartIndex() method, so null won't be reached here, is there any reason to do it here again?

somindatommy: This is a fix asked in a previous commit as a precaution.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1697,10 +1734,29 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
...
+    @Override
+    public List<Object> listGroupsWithGET(Node rootNode, Integer startIndex, Integer count, String sortBy,
+            String sortOrder, String domainName, Map<String, Boolean> requiredAttributes)
+            throws CharonException, NotImplementedException {
+
+        // Validate NULL value for startIndex.
+        startIndex = handleStartIndexEqualsNULL(startIndex);
+        if (sortBy != null || sortOrder != null) {
+            throw new NotImplementedException("Sorting is not supported");
+        } else if (startIndex != 1 || count != null) {
```
malithie: Do we need to handle null start index above.
Simple it's not null means and in input is provided right.
So just mentioning it's not supported if it's not null is enough right

malithie: Meanwhile it's consistent with other arguments being used here and more readable

somindatommy: according to the spec, start Index param is optional. In that case we have to set it to 1 as it is in the specification. So mentioning as "Not Supported" is not an option.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1697,10 +1734,29 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
     @Override
     public List<Object> listGroupsWithGET(Node rootNode, int startIndex, int count, String sortBy, String sortOrder,
             String domainName, Map<String, Boolean> requiredAttributes)
+
             throws CharonException, NotImplementedException, BadRequestException {
         if (sortBy != null || sortOrder != null) {
```
tharindu-bandara: Move above new line to here, where the method starts

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
tharindu-bandara: But we are not implementing pagination support for `listGroupsWithGET` right? Then For this implementation what needs to be returned is pagination is not supported, if user request comes with a start index. Even though you handle startIndex correctly, it won't be used unless pagination is properly implemented.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: Fixed: 2aa3adf18f5d3517e09c218c2c8be0eb29ac6e4e

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1709,6 +1765,24 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
...
+     * Method to interpret startIndex as 1 when the startIndex equals to NULL in the request.
+     *
+     * @param startIndex StartIndex in the request.
+     * @return Updated startIndex
+     */
+    private int handleStartIndexEqualsNULL(Integer startIndex) {
+
+        if (startIndex == null) {
+            if (log.isDebugEnabled()) {
+                log.debug("NULL argument for startIndex in the request. NULL value interpreted as 1 and updated "
```
tharindu-bandara: Shall we summarize this debug log to reflect the operation, to something like `NULL value returned for the startIndex argument interpreted as 1`?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: Refer to line 1758. If the startIndex value is not equal to 1, it throws an exception. The Null implementation for startIndex is correct because returning 1 is the default case.

components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
```diff
@@ -312,59 +312,76 @@ private Response processRequest(final Map<String, String> requestAttributes) {
...
         } catch (CharonException e) {
             return handleCharonException(e, encoder);
         }
     }
 
+    /**
+     * Method to convert string pagination values to Interger pagination values.
+     *
+     * @param valueInRequest       Value passed in the request.
+     * @param scimProviderConstant The name of parameter (count or startIndex).
```
tharindu-bandara: Shall we remove `count or startIndex` ?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
tharindu-bandara: Ok, I see the point. This way both `null` and `1` cases are handled, and anything else means the pagination which we don't support.

components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/GroupResource.java
somindatommy: Fixed: 2a9df39b802d0a818f4c26bdd0f6c7ffe48a4136

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
somindatommy: fixed: 2a9df39b802d0a818f4c26bdd0f6c7ffe48a4136

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
malithie: +1

PR #165, opened 2019-04-30
Fix SCIMCommonUtils.isFilteringEnhancementsEnabled() property is used in both request flow and response flow for two different behaviours

IshikaGodage:
Fix [wso2/product-is #4961](https://github.com/wso2/product-is/issues/4961)

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
@@ -67,6 +67,7 @@ public class GroupDAO {
             while (resultSet.next()) {
                 String group = resultSet.getString(1);
                 if (StringUtils.isNotEmpty(group)) {
+                    group = SCIMCommonUtils.getPrimaryFreeGroupName(group);
                     groups.add(group);
                 }
             }
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -81,6 +81,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils.mandateDomainForGroupNamesInGroupsResponse;
+import static org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils.mandateDomainForUsernamesAndGroupNamesInResponse;
+import static org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils.isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled;
+import static org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils.isFilteringEnhancementsEnabled;
+import static org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils.prependDomain;
+
 public class SCIMUserManager implements UserManager {
 
     public static final String FILTERING_DELIMITER = "*";
@@ -706,17 +712,7 @@ public class SCIMUserManager implements UserManager {
                     node.getValue()));
         }
 
-        // Apply filter enhancements for username and update expression node value.
-        applyFilterEnhancementsForUsernameInNode(node);
-        try {
-            // Extract he domain name if the domain name is embedded in the filter attribute value.
-            domainName = resolveDomainNameInAttributeValue(domainName, node);
-        } catch (BadRequestException e) {
-            String errorMessage = String
-                    .format("Domain parameter: %s in request does not match with the domain name in the attribute "
-                            + "value: %s ", domainName, node.getValue());
-            throw new CharonException(errorMessage, e);
-        }
+        domainName = resolveDomainName(domainName, node);
         try {
             // Check which APIs should the filter needs to follow.
             if (isUseLegacyAPIs(limit)) {
@@ -729,6 +725,32 @@ public class SCIMUserManager implements UserManager {
             throw new CharonException(errorMessage, e);
         }
         return getDetailedUsers(userNames, requiredAttributes);
+    }
+
+    /**
+     * Method to resolve the domain name.
+     *
+     * @param domainName Domain to run the filter
+     * @param node       Expression node for single attribute filtering
+     * @return Resolved domainName
+     * @throws CharonException
+     */
+    private String resolveDomainName(String domainName, ExpressionNode node) throws CharonException {
+
+        try {
+            // Extract the domain name if the domain name is embedded in the filter attribute value.
+            domainName = resolveDomainNameInAttributeValue(domainName, node);
+        } catch (BadRequestException e) {
+            String errorMessage = String
+                    .format("Domain parameter: %s in request does not match with the domain name in the attribute "
+                                    + "value: %s ", domainName, node.getValue());
+            throw new CharonException(errorMessage, e);
+        }
+        // Get domain name according to Filter Enhancements properties as in identity.xml
+        if (StringUtils.isEmpty(domainName)) {
+            domainName = getFilteredDomainName(node);
+        }
+        return domainName;
     }
 
     /**
@@ -751,25 +773,32 @@ public class SCIMUserManager implements UserManager {
     }
 
     /**
-     * Validate whether filter enhancements are enabled and then append the primary domain name in front of the
-     * attribute value to be searched. Finally update the attribute value in Expression node to new attribute value.
+     * Validate whether filter enhancements are enabled and then return primary default domain name as the domain to
+     * be filtered.
      *
      * @param node Expression node
-     * @return Modified attribute value
+     * @return PRIMARY domainName if property enabled, Null otherwise.
      */
-    private void applyFilterEnhancementsForUsernameInNode(ExpressionNode node) {
+    private String getFilteredDomainName(ExpressionNode node) {
 
         // Set filter values.
         String attributeName = node.getAttributeValue();
         String filterOperation = node.getOperation();
         String attributeValue = node.getValue();
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled())
-            if (SCIMCommonConstants.EQ.equalsIgnoreCase(filterOperation))
+        if ((isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) && !StringUtils
+                .contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)) {
+            return UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME;
+
+        } else if (isFilteringEnhancementsEnabled()) {
+            if (SCIMCommonConstants.EQ.equalsIgnoreCase(filterOperation)) {
                 if (StringUtils.equals(attributeName, SCIMConstants.UserSchemaConstants.USER_NAME_URI) && !StringUtils
-                        .contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR))
-                    node.setValue(UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME + CarbonConstants.DOMAIN_SEPARATOR
-                            + attributeValue);
+                        .contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)) {
+                    return UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME;
+                }
+            }
+        }
+        return null;
     }
 
     /**
@@ -1763,12 +1792,12 @@ public class SCIMUserManager implements UserManager {
             throw new NotImplementedException(error);
         }
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled()) {
+        if (isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) {
+            attributeValue = prependDomain(attributeValue);
+        } else if (isFilteringEnhancementsEnabled()) {
             if (SCIMCommonConstants.EQ.equalsIgnoreCase(filterOperation)) {
-                if (StringUtils.equals(attributeName, SCIMConstants.GroupSchemaConstants.DISPLAY_NAME_URI) &&
-                        !StringUtils.contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)) {
-                    attributeValue = UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME +
-                            CarbonConstants.DOMAIN_SEPARATOR + attributeValue;
+                if (StringUtils.equals(attributeName, SCIMConstants.GroupSchemaConstants.DISPLAY_NAME_URI)) {
+                    attributeValue = prependDomain(attributeValue);
                 }
             }
         }
@@ -1790,22 +1819,13 @@ public class SCIMUserManager implements UserManager {
                                 UserCoreUtil.isEveryoneRole(roleName, carbonUM.getRealmConfiguration())) {
                             continue;
                         }
-                        /**construct the group name with domain -if not already provided, in order to support
-                         multiple user store feature with SCIM.**/
-                        String groupNameWithDomain = null;
-                        if (roleName.indexOf(CarbonConstants.DOMAIN_SEPARATOR) > 0) {
-                            groupNameWithDomain = roleName;
-                        } else {
-                            groupNameWithDomain = UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME + CarbonConstants.DOMAIN_SEPARATOR
-                                    + roleName;
-                        }
                         String userStoreDomainName = IdentityUtil.extractDomainFromName(roleName);
                         if (isInternalOrApplicationGroup(userStoreDomainName) || isSCIMEnabled(userStoreDomainName)) {
                             if (log.isDebugEnabled()) {
                                 log.debug("SCIM is enabled for the user-store domain : " + userStoreDomainName + ". " +
                                         "Including group with name : " + roleName + " in the response.");
                             }
-                            Group group = getGroupWithName(groupNameWithDomain);
+                            Group group = getGroupWithName(roleName);
                             filteredGroups.add(group);
                         } else {
                             if (log.isDebugEnabled()) {
@@ -2070,6 +2090,10 @@ public class SCIMUserManager implements UserManager {
 
             //get groups of user and add it as groups attribute
             String[] roles = carbonUM.getRoleListOfUser(userName);
+            // Add username with domain name
+            if (mandateDomainForUsernamesAndGroupNamesInResponse()) {
+                userName = prependDomain(userName);
+            }
             //construct the SCIM Object from the attributes
             scimUser = (User) AttributeMapper.constructSCIMObjectFromAttributes(attributes, 1);
 
@@ -2086,10 +2110,10 @@ public class SCIMUserManager implements UserManager {
                     continue;
                 }
 
-                if (SCIMCommonUtils.isFilteringEnhancementsEnabled()) {
-                    if (!StringUtils.contains(role, CarbonConstants.DOMAIN_SEPARATOR)) {
-                        role = UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME + CarbonConstants.DOMAIN_SEPARATOR + role;
-                    }
+                if (mandateDomainForUsernamesAndGroupNamesInResponse()) {
+                    role = prependDomain(role);
+                } else if (isFilteringEnhancementsEnabled()) {
+                    role = prependDomain(role);
                 }
 
                 Group group = groupMetaAttributesCache.get(role);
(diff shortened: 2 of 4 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
```diff
@@ -67,6 +67,9 @@
             while (resultSet.next()) {
                 String group = resultSet.getString(1);
                 if (StringUtils.isNotEmpty(group)) {
+                    if (StringUtils.isNotEmpty(group)) {
```
malithie: Isn't the same condition duplicated here

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -731,6 +727,31 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
         return getDetailedUsers(userNames, requiredAttributes);
     }
 
+    /**
+     *
```
Abilashini: Please remove the additional line

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -731,6 +727,31 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
...
+     * Method to resolve the domain name.
+     * @param domainName Domain to run the filter
+     * @param node       Expression node for single attribute filtering
+     * @return Resolved domainName
+     * @throws CharonException
+     */
+    private String getDomainName(String domainName, ExpressionNode node) throws CharonException {
+        try {
```
Abilashini: There should be a new line after the method signature

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -751,25 +772,33 @@ private boolean isUseLegacyAPIs(int limit) {
     }
 
     /**
-     * Validate whether filter enhancements are enabled and then append the primary domain name in front of the
-     * attribute value to be searched. Finally update the attribute value in Expression node to new attribute value.
+     *
```
Abilashini: Remove the additional line

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2070,6 +2090,10 @@ private User getSCIMUser(String userName, List<String> claimURIList, Map<String,
 
             //get groups of user and add it as groups attribute
             String[] roles = carbonUM.getRoleListOfUser(userName);
+            // Add username with domain name
+            if (MandateDomainForUsernamesAndGroupNamesInResponse()) {
```
malithie: Start a method with a simple letter

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -304,4 +304,54 @@ public static boolean isFilteringEnhancementsEnabled() {
...
+                .getProperty(SCIMCommonConstants.SCIM_ENABLE_FILTER_USERS_AND_GROUPS_ONLY_FROM_PRIMARY_DOMAIN));
+    }
+
+    /**
+     * Checks whether the identity.xml config is available for prepending the 'PRIMARY/' in each role and
+     * which belong to Primary domain in the response of Groups endpoints.
+     *
+     * @return whether 'MandateDomainForGroupNamesInGroupsResponse' property is enabled in identity.xml.
+     */
+    public static boolean MandateDomainForGroupNamesInGroupsResponse() {
```
malithie: start a method with a simple letter

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1763,12 +1792,12 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
             throw new NotImplementedException(error);
         }
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled()) {
+        if (isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) {
+            attributeValue = prependPrimaryDomain(attributeValue);
```
malithie: This adds primary domain to every attribute. Is this necessary. Only specific group attributes could have domain prepended and handled right similar to username

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -751,25 +773,32 @@ private boolean isUseLegacyAPIs(int limit) {
...
 
         // Set filter values.
         String attributeName = node.getAttributeValue();
         String filterOperation = node.getOperation();
         String attributeValue = node.getValue();
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled())
-            if (SCIMCommonConstants.EQ.equalsIgnoreCase(filterOperation))
+        if (isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) {
+            if (!StringUtils.contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)) {
```
tharindu-bandara: Shall we combine the two conditions `isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()` and `!StringUtils.contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)` and use a single `if condition` instead of two to improve readability? If the combined condition seem long and complex, you can wrap it in a private method with a suitable name reflecting the condition it checks.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Followings are the supported attributes in Groups endpoint.

- displayName
- members
- members.value
- members.$ref
- members.display
- members.type

Previously we add the 'PRIMARY/' in-front of the 'displayName' attribute only. However when newly introduced property (FilterUsersAndGroupsOnlyFromPrimaryDomain) enabled, The 'PRIMARY/' will be added in-front of every attribute which are listed above.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -751,25 +773,32 @@ private boolean isUseLegacyAPIs(int limit) {
...
+    private String getFilteredDomainName(ExpressionNode node) {
 
         // Set filter values.
         String attributeName = node.getAttributeValue();
         String filterOperation = node.getOperation();
         String attributeValue = node.getValue();
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled())
-            if (SCIMCommonConstants.EQ.equalsIgnoreCase(filterOperation))
+        if ((isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) && !StringUtils
```
tharindu-bandara: Shall we remove the redundant parenthesises?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -751,25 +773,32 @@ private boolean isUseLegacyAPIs(int limit) {
...
+                .contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)) {
+            return UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME;
+
```
tharindu-bandara: Shall we remove this new line?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
tharindu-bandara: Both conditions in `if` and `else if` checks for the two scenarios, to return the primary default domain name. Shall we use a single private method, which checks if the current scenario is equal to either of the two scenarios? This will reduce the code essentially to a single if condition to return primary default domain name and return null if not.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -731,6 +727,32 @@ public User updateUser(User user, Map<String, Boolean> requiredAttributes) throw
...
 
+    /**
+     * Method to resolve the domain name.
+     *
+     * @param domainName Domain to run the filter
+     * @param node       Expression node for single attribute filtering
+     * @return Resolved domainName
+     * @throws CharonException
+     */
+    private String getDomainName(String domainName, ExpressionNode node) throws CharonException {
```
tharindu-bandara: Shall we rename this method to `resolveDomainName` to reflect it's functionality?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
tharindu-bandara: Does `prependPrimaryDomain` method handle the scenario where `attributeValue` already comes with a domain prepended?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1763,12 +1792,12 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
             throw new NotImplementedException(error);
         }
 
-        if (SCIMCommonUtils.isFilteringEnhancementsEnabled()) {
+        if (isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) {
```
tharindu-bandara: This is checking for two scenarios to prepend the primary domain. Shall we follow a similar approach to https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/165/files#r279851284

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1790,22 +1819,13 @@ public void deleteGroup(String groupId) throws NotFoundException, CharonExceptio
...
-                                    + roleName;
-                        }
                         String userStoreDomainName = IdentityUtil.extractDomainFromName(roleName);
                         if (isInternalOrApplicationGroup(userStoreDomainName) || isSCIMEnabled(userStoreDomainName)) {
                             if (log.isDebugEnabled()) {
                                 log.debug("SCIM is enabled for the user-store domain : " + userStoreDomainName + ". " +
                                         "Including group with name : " + roleName + " in the response.");
                             }
-                            Group group = getGroupWithName(groupNameWithDomain);
+                            Group group = getGroupWithName(roleName);
```
tharindu-bandara: From where does the `roleName` is prepended with the domain?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2276,13 +2301,24 @@ private Group getGroupWithName(String groupName)
         }
 
         Group group = new Group();
-        group.setDisplayName(groupName);
+        if (mandateDomainForUsernamesAndGroupNamesInResponse()) {
+            groupName = prependPrimaryDomain(groupName);
+            group.setDisplayName(groupName);
+        } else if (mandateDomainForGroupNamesInGroupsResponse()) {
```
tharindu-bandara: In what kind of a scenario, this condition and the previous `if` condition differ?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -304,4 +304,54 @@ public static boolean isFilteringEnhancementsEnabled() {
         return Boolean.parseBoolean(IdentityUtil.getProperty
                 (SCIMCommonConstants.SCIM_ENABLE_FILTERING_ENHANCEMENTS));
     }
+
+    /**
+     * Checks whether the identity.xml config is available for applying filtering enhancements. This makes sure that Eq
+     * should strictly check for the String match (in this case cross user store search should not be performed).
```
tharindu-bandara: Shall we simplify this method description a bit? the second sentence looks a bit complex to understand. This config will filter only from the primary domain if the domain is not explicitly present?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -304,4 +304,54 @@ public static boolean isFilteringEnhancementsEnabled() {
...
+        return Boolean.parseBoolean(IdentityUtil
+                .getProperty(SCIMCommonConstants.SCIM_ENABLE_MANDATE_DOMAIN_FOR_USERNAMES_AND_GROUPNAMES_IN_RESPONSE));
+    }
+
+    /**
+     * Method to prepend PRIMARY domain name to given String when that String not contains DOMAIN_SEPARATOR.
+     * @param value
+     * @return Given String if given String not contains DOMAIN_SEPARATOR, String with PRIMARY domain otherwise.
+     */
+    public static String prependPrimaryDomain(String value) {
```
tharindu-bandara: Shall we change the method name to reflect the internals? This name does not say that the `PRIMARY` would not be appended if already a domain name is present

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
malithie: My question is why should it add.
domain separator can be a a character used in any other attribute that displayName and members.value.
How in such a case primary domain would be identified
The intention here is, if not domain specified (in an attribute where user store domain is allowed like displayName or as the 'domain' query param) to switch to filter only from primary domain.
The method used here is error prone, moreover, from each such attribute when filtering from the user store it self, is this prepended domain handled

IshikaGodage: @tharindu-bandara yes, `prependPrimaryDomain` method handle the scenario where attributeValue already comes with a domain prepended. I'll change the method name.

@malithie, Currently domain  query param is not supported. Therefore, I'll  change the above to prepend 'PRIMARY/' only for  **displayName** and **members.value**. Once domain query param support for Groups endpoint. We can do the same thing what we did as Users endpoint.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: If the role not belongs to PRIMARY userstore, PRIMARY is not in the roleName. If we enable the properties in the identity.xml, Based on the properties PRIMARY will be prepended in the `getGroupWithName` method.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
IshikaGodage: There was a case where primary domain is not depicted with users being returned at all but only on groups.We discussed to have a separate property to enable that behavior to ensure backward compatibility and to deprecate that in a later version. This is to check that separate property.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
IshikaGodage: Fixed

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
malithie: I still don't think this sorts out the problem. Also members.value seems to refer to the user id and not user name. Then again it should be members.display
Can you check the attributes and fix accordingly.
Anyhow irrespective of the attribute you have to make sure if it's set to list or filter from primary to filter only from primary.
Basically, if there's no domain specified in the domain allowed attributes, filtering from any attribute should happen from the primary domain. IMO, for groups may be the API does not support domain, but internally you can resolve the domain to list of filter from and you can list or filter from the respective user store domain

IshikaGodage: Currently Groups endpoint not supported for domain query param. Once it is done by this PR[1],
It is possible to resolve the domain name and use the existing methods as we did in the Users endpoint.
[1][https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/166](https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/166)

PR #189, opened 2019-08-08
Support hybrid role listing & filtering with domain parameter for SCIM

denuwanthi:
Resolves wso2/product-is#5954

In this PR it supports hybrid roles (Application &Internal) listing and filtering.
Listing can be done using a new 'domain' query param.
Filtering can be done in 2 ways.
1.using 'domain' query param
2.using '{domain}/' prefix

Example listing request:
curl -v -k --user admin:admin 'https://localhost:9443/scim2/Groups?domain=Application'

Example filtering request with domain parameter:
curl -v -k --user admin:admin 'https://localhost:9443/scim2/Groups?filter=displayName+eq+Apple&domain=Internal'

Example filtering request with domain prefix:
curl -v -k --user admin:admin 'https://localhost:9443/scim2/Groups?filter=displayName+sw+Application/App'

PR #190, opened 2019-08-12
Support hybrid role listing & filtering with domain parameter for SCIM

denuwanthi:
Resolves wso2/product-is#5954

In this PR it supports hybrid roles (Application &Internal) listing and filtering.
Listing can be done using a new 'domain' query param.
Filtering can be done in 2 ways.
1.using 'domain' query param
2.using '{domain}/' prefix

Example listing request:
curl -v -k --user admin:admin 'https://localhost:9443/scim2/Groups?domain=Application'

Example filtering request with domain parameter:
curl -v -k --user admin:admin 'https://localhost:9443/scim2/Groups?filter=displayName+eq+Apple&domain=Internal'

Example filtering request with domain prefix:
curl -v -k --user admin:admin 'https://localhost:9443/scim2/Groups?filter=displayName+sw+Application/App'


denuwanthi:
This PR was initially in https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/189
Due to branch switch now sent the same PR to master

PR #204, opened 2019-09-20
Updated scim2.yaml with swagger

GDRDABARERA:


janakamarasena:
Shall we change email: "architecture@wso2.com" to email: "architecture@wso2.org" as well

scim2.yaml
```diff
@@ -84,28 +97,48 @@ paths:
...
+          description: Internal Server Error
+          schema:
+            $ref: '#/definitions/ErrorInternalServerError'
+
     put:
       tags:
         - Me Endpoint
-      summary: Update Me - PUT 
-      description: This API uses a PUT operation to update user details. It returns HTTP 404 if the user is not found.
+      summary: Update Me - PUT
```
janakamarasena: Mentioning the operation type (PUT) in here and the description is not needed.

scim2.yaml
```diff
@@ -123,50 +156,63 @@ paths:
...
+          schema:
+            $ref: '#/definitions/ErrorNotimplemented'
     patch:
       tags:
         - Me Endpoint
-      summary: Update Me - PATCH 
-      description: This API uses a PATCH operation to update user details. It returns HTTP 404 if the user is not found.
+      summary: Update Me - PATCH
+      description: |
+        This API uses a PATCH operation to update user details. It returns HTTP 404 if the user is not found.
```
janakamarasena: Since the operation and the response codes are defined we can omit these from the description and describe what happens when the API is invoked

PR #251, opened 2020-02-26
Remove unnecessary logs

nilasini:
Resolves https://github.com/wso2/product-is/issues/6053


darshanasbg:
Are these the only places that have the info logs?

Can we mark them as debug or trace, rather completely removing them?

nilasini:
We have info logs in some other places as well but that is for deletion and update. I thought those info logs would be helpful to have it as an info log. Should I change those as a debug log?

When we try to get a user if the user is not there then we are printing it in a debug log. Do we still need to print a debug log if the user is there as well? Won't this make a lot of logs in the console?

darshanasbg:
I think we have to change others to debug as well.. Tracking shouldn't be a problem as we have audit logs to track any profile updates. Having info logs again in scim level is redundant IMO.

And for the second query, its ok to have them as debug logs as those won't be enabled by default or in usual cases. Those will be needed to turned on, only when isolating some issue. Hence any log that would help to debug and give that level information, is ok to keep as debug logs.

nilasini:
Ok, thank you for your clarifications. I will send another PR with the suggested changes.

PR #263, opened 2020-05-26
Build SCIM2 endpoint URLs and resolve tenant domain for tenant context mode

Abilashini:
Related to https://github.com/wso2/product-is/issues/8404

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/IdentitySCIMManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/IdentitySCIMManager.java
@@ -22,6 +22,8 @@ import org.apache.commons.collections.MapUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.context.PrivilegedCarbonContext;
+import org.wso2.carbon.identity.core.util.IdentityTenantUtil;
+import org.wso2.carbon.identity.scim2.common.exceptions.IdentitySCIMException;
 import org.wso2.carbon.identity.scim2.common.internal.SCIMCommonComponentHolder;
 import org.wso2.carbon.identity.scim2.common.utils.AuthenticationSchema;
 import org.wso2.carbon.identity.scim2.common.utils.SCIMCommonConstants;
@@ -29,11 +31,8 @@ import org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils;
 import org.wso2.carbon.identity.scim2.common.utils.SCIMConfigProcessor;
 import org.wso2.carbon.user.api.UserRealm;
 import org.wso2.carbon.user.api.UserStoreException;
-import org.wso2.carbon.user.core.UserStoreManager;
-import org.wso2.carbon.user.core.claim.ClaimManager;
 import org.wso2.carbon.user.core.common.AbstractUserStoreManager;
 import org.wso2.carbon.user.core.service.RealmService;
-import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 import org.wso2.charon3.core.config.CharonConfiguration;
 import org.wso2.charon3.core.encoder.JSONEncoder;
 import org.wso2.charon3.core.exceptions.CharonException;
@@ -66,6 +65,9 @@ public class IdentitySCIMManager {
      */
     public static IdentitySCIMManager getInstance() throws CharonException {
 
+        if (IdentityTenantUtil.isTenantQualifiedUrlsEnabled()) {
+            return new IdentitySCIMManager();
+        }
         if (identitySCIMManager == null) {
             synchronized (IdentitySCIMManager.class) {
                 if (identitySCIMManager == null) {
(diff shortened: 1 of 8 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/IdentitySCIMManager.java
```diff
@@ -66,6 +65,9 @@ private IdentitySCIMManager() throws CharonException {
      */
     public static IdentitySCIMManager getInstance() throws CharonException {
 
+        if (IdentityTenantUtil.isTenantQualifiedUrlsEnabled()) {
+            return new IdentitySCIMManager();
+        }
```
ashensw, line 70: Any reason to change this method?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -66,39 +70,54 @@
...
+            }
         } else {
-            scimURL = IdentityUtil.getServerURL(SCIMCommonConstants.SCIM2_ENDPOINT, true, true);
+            String tenantDomain = PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain();
+            if (isNotASuperTenantFlow(tenantDomain)) {
+                scimURL = IdentityUtil.getServerURL(SCIMCommonConstants.TENANT_URL_SEPERATOR + tenantDomain +
+                        SCIMCommonConstants.SCIM2_ENDPOINT, true, true);
+            } else {
+                scimURL = IdentityUtil.getServerURL(SCIMCommonConstants.SCIM2_ENDPOINT, true, true);
+            }
```
ashensw, line 120: Do we need this part? Inside the ServiceURLBuilder also it's checking the `isTenantQualifiedUrlsEnabled()` config. So if there is no additional config file used we don't need to specifically check the `isTenantQualifiedUrlsEnabled()` again here. If so we can get rid of the old `IdentityUtil.getServerURL()` completely.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -107,16 +126,38 @@ private static boolean isNotASuperTenantFlow(String tenantDomain) {
...
+            try {
+                scimURL = ServiceURLBuilder.create().addPath(SCIMCommonConstants.SCIM2_ENDPOINT).build()
+                        .getAbsolutePublicURL();
+            } catch (URLBuilderException e) {
+                throw new IdentitySCIMException("Error occurred while building the SCIM2 endpoint with tenant " +
+                        "qualified URL.", e);
+            }
+        } else {
+            scimURL = IdentityUtil.getServerURL(SCIMCommonConstants.SCIM2_ENDPOINT, true, true);
+        }
```
ashensw, line 142: Same as the above comment.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -107,16 +126,38 @@ private static boolean isNotASuperTenantFlow(String tenantDomain) {
...
```
ashensw, line 159: Same as the above comment.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/IdentitySCIMManager.java
Abilashini: IdentitySCIMManager object is instantiated during server startup and set the endpoint URLs in AbstractResourceManager with the super tenant URLs at [1]. AbstractResourceManager is accessed by charon to get the endpoint URLs. So for tenants as well the endpoint URLs will the super tenant URLs. 
[1] https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/263/files/42a6b93ef0c49de05581e2fe2dc57dd4a3820aea#diff-609fb5ae29747e7d7564f70cbaa60e44R105

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
Abilashini: If the configuration is disabled ServiceURLBuilder doesn't return the tenant domain

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/IdentitySCIMManager.java
emswbandara: this will instantiate a new instance of IdentitySCIMManager, everytime getInstance() is called. Better to have a instance variable and instantiate only if that variable is null.

components/org.wso2.carbon.identity.scim2.provider/src/main/java/org/wso2/carbon/identity/scim2/provider/resources/BulkResource.java
```diff
@@ -66,7 +66,7 @@ public Response createUser(@HeaderParam(SCIMProviderConstants.AUTHORIZATION) Str
             encoder = identitySCIMManager.getEncoder();
 
             // obtain the user store manager
-            UserManager userManager = IdentitySCIMManager.getInstance().getUserManager();
+            UserManager userManager = identitySCIMManager.getUserManager();
```
emswbandara, line 69: why was this change required?

Abilashini: To avoid instantiating IdentitySCIMManager twice in the same method

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/IdentitySCIMManager.java
Abilashini: We should instantiate IdentitySCIMManager every time if we want to have tenant qualified URLs. 
Anyways we will be doing this only if tenant context is enabled. otherwise, it will be instantiated only once and the same instance will be used for the follow-up calls.

malithie: We should not re-instantiate. Rather we should fix getters or add getters if not available to build url based on context

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
malithie: So in the legacy way, we still had the tenant qualified complete url built ?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
```diff
@@ -177,12 +218,22 @@ public static void setThreadLocalIsManagedThroughSCIMEP(Boolean value) {
     }
 
     public static String getGlobalConsumerId() {
+
+        if (IdentityTenantUtil.isTenantQualifiedUrlsEnabled()) {
+            return IdentityTenantUtil.getTenantDomainFromContext();
```
malithie, line 223: Wondering whether it's good to return tenant domain from privileged carbon context if thread local is not available

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/SCIMCommonUtils.java
Abilashini: yes

Abilashini:
Re-initiated and fixed at https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/265

PR #275, opened 2020-08-09
Fix alignment issue in scim2 configs

pamodaaw:


pamodaaw:
Another PR is made for this change https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/276#pullrequestreview-463847597

Issue #271, opened 2020-07-23
Hardcoded policy errors


geve82:
**Description:**
Policy errors list is hardcoded in org.wso2.carbon.identity.scim2.common.impl.SCIMUserManager.handleErrorsOnUserNameAndPasswordPolicy(Throwable e).

Natively all username and password code are not managed.
Developers can also add their own custom policies and errors code.

If an unknown error code is received a generic CharonException is thrown and a 500 is returned to the caller. 
`{"schemas":["urn:ietf:params:scim:api:messages:2.0:Error"],"detail":"Error while updating attributes of user: test","status":"500"}`
But in those cases it's a user mistake so a 4XX must be returned and an explicit error message must be sent back depending of the policy.

**Suggestion:**
Create an error code range convention for user errors or a specific class exception to catch.

**Suggested Labels:**
BUG

**Affected Product Version:**
All

**OS, DB, other environment details and versions:**    
N/A

**Steps to reproduce:**
Create your own password policy and try a bad pattern.

**Related Issues:**
N/A

ruwanta:
There are few error codes and types defined in SCIM specification [1]
What we can do is to have a error translator, to translate the given exception or error to the standard defined by SCIM.

SCIM allows the error detail to be more verbose and human readable. 

So we could think of having translation mechanism to {ErrorCode, DcimType, Detail} tuple from internal exception or error

[1] https://tools.ietf.org/html/rfc7644#page-67

PR #288, opened 2020-09-16
Exception handling for IdentityEventException

astik:
Follow up of #271, this PR introduces a new settings category: IdentityEventExceptionSettings which contains 2 parameters:
    - exposeErrorCodeInMessage: whether or not we should expose the internal IdentityEventExceptionSettings error code
    - badRequestErrorCodes: list of error code that should trigger a BadRequestException

In current codebase, an IdentityEventException in SCIMUserManager is transformed into a BadRequestException if the error code is "2201" (which will trigger a 400 error code). If not, it is wrapped in a CharonException (which will trigger a 500 error code).

In some use cases, we might need other error code to be transformed to bad request: this is what *badRequestErrorCodes* is for.
Also in some cases, we might need some additional context to personalized message in the UI. Without any help, we are doomed to an error message coming from the Java exception itself. Having access to the error code in the message is a good enough solution: this is what *exposeErrorCodeInMessage* is for.

I also add some unit test:
- first to cover the existing use case which was not tested
- then to cover the changes in the settings loading and in the SCIMUserManager updated exception handler.

What do you thing about this approach ?

(there is still some polish to do in the SCIMUserManager class as there are elements which are not used (dead code): ROLE_CLAIM, doUserValidation, addDomainToUserMembers, getMappedClaimList and now ERROR_CODE_PASSWORD_HISTORY_VIOLATION).

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -48,8 +48,10 @@ import org.wso2.carbon.identity.scim2.common.extenstion.SCIMUserStoreException;
 import org.wso2.carbon.identity.scim2.common.group.SCIMGroupHandler;
 import org.wso2.carbon.identity.scim2.common.internal.SCIMCommonComponentHolder;
 import org.wso2.carbon.identity.scim2.common.utils.AttributeMapper;
+import org.wso2.carbon.identity.scim2.common.utils.IdentityEventExceptionSettings;
 import org.wso2.carbon.identity.scim2.common.utils.SCIMCommonConstants;
 import org.wso2.carbon.identity.scim2.common.utils.SCIMCommonUtils;
+import org.wso2.carbon.identity.scim2.common.utils.SCIMConfigProcessor;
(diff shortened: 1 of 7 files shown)


astik:
Some new stuff has been merged for exception handling:

- 331f316496fe94f62194a1a1edf0d0460b245e17 : adding BadRequestException everywhere in SCIMUserManager
- bb72f813d9867fc8df43f9a40c89b80391f67cee : introducing SCIMUserStoreErrorResolver.java

@geve82 maybe the SCIMUserStoreErrorResolver might be what you were looking for when you ask me for this PR?
Is the PR still relevant with the newly upstream code? Should we rebase? should we give up this one?

madurangasiriwardena:
Hi @astik,

Thank you for your contribution and sorry about the delay in reviewing the PR.

As you mentioned I believe SCIMUserStoreErrorResolver introduced [here](https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/311) full fill your requirement. I believe it will fit better for general use cases since the error code mapping are not required to be changed frequently (with configurations) and associates with the code developed for user store managers and event handlers in user/group management flows. Hence implementing an error resolve would be better to full fill the requirements.

Please let me know your thoughts.

Regards,
Maduranga.

madurangasiriwardena:
Hi @astik,

Based on my previous comment, I'm closing this PR. Please feel free to reopen if you think otherwise.

Thank you for your contribution and looking forward for more contributions in the future.

Regards,
Maduranga.

PR #307, opened 2020-11-11
Revert "Sub attribute names made unique in the scim2-schema-extension.config"

ashendes:
Reverting this as this would change the format of the request payloads.


IsuraD (approved):

piraveena:
Reverted https://github.com/wso2/product-is/issues/9548

PR #321, opened 2021-01-07
Fix the issue with group patch operation when using email as username

nilasini:
Resolves https://github.com/wso2/product-is/issues/9800


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2896,6 +2897,10 @@ public Group patchGroup(String groupId, String currentGroupName, Map<String, Lis
                 validateUserIds(addedMemberIdsFromUserstore, newlyAddedMemberIds);
             }
 
+            if (isNotEmpty(deletedMembers) && isNotEmpty(deletedMemberIds)) {
```
IsuraD, line 2900: do we need to check isNotEmpty(deletedMembers) here?

IsuraD (approved):

PR #322, opened 2021-01-07
Remove member attribute from the group response

nilasini:
Resolves https://github.com/wso2/product-is/issues/9619
If the member attribute doesn't have the value then the member will not be added to the group endpoint hence removing it from the response.


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2187,6 +2187,7 @@ public Group createGroup(Group group, Map<String, Boolean> requiredAttributes)
                 SCIMGroupHandler scimGroupHandler = new SCIMGroupHandler(carbonUM.getTenantId());
                 scimGroupHandler.createSCIMAttributes(group);
                 carbonUM.addRoleWithID(group.getDisplayName(), null, null, false);
+                group.getAttributeList().remove(SCIMConstants.GroupSchemaConstants.MEMBERS);
```
IsuraD, line 2190: why do we remove member attribute ? Is that according to the spec?

PR #338, opened 2021-03-11
Fix the case insensitivity issue when updating the group name

GANGANI:
## Purpose
When executing a SCIM group update request to update the same group name with case sensitivity(Ex:- update mygroup to MyGroup), it will give a log warn saying "There is no updated field in the group:... " and will not update the group name. That issue will be fixed with this PR. 
Resolves: https://github.com/wso2-enterprise/asgardeo-product/issues/1655

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -3369,7 +3369,7 @@ public class SCIMUserManager implements UserManager {
 
     private boolean isGroupDisplayNameChanged(String oldGroupDisplayName, String newGroupDisplayName) {
 
-        return !oldGroupDisplayName.equalsIgnoreCase(newGroupDisplayName);
+        return !oldGroupDisplayName.equals(newGroupDisplayName);
     }
 
     private org.wso2.carbon.user.core.common.User getUserFromUsername(String username)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -3369,7 +3369,7 @@ private boolean isPrimaryDomain(String newGroupDomainName) {
 
     private boolean isGroupDisplayNameChanged(String oldGroupDisplayName, String newGroupDisplayName) {
 
-        return !oldGroupDisplayName.equalsIgnoreCase(newGroupDisplayName);
+        return !oldGroupDisplayName.equals(newGroupDisplayName);
```
tharindu-bandara, line 3372: Does this method only used in `doUpdateGroup `? If so, from where it says Group already exists error as if this method says group name is not changed, it does not perform a DB check, correct?

GANGANI: Yes. This method is only used in `doUpdateGroup`. It doesn't give any group already exist error. But it will not update the DB with a log warn saying "There is no updated field in the group:"

tharindu-bandara: Got it thanks! Thought the issue mentioned in the PR description is directly related. Shall we update the description mentioning what would be the exact thing this PR is trying to fix? I think the mentioned issue is for a more general case and does not mention the part that we are fixing here

tharindu-bandara (approved):

IsuraD (approved):

PR #433, opened 2022-07-07
Modify API response URLs for organization specific routing scenarios

dewniMW:
Issue: https://github.com/wso2/product-is/issues/14060

Please note that location header URIs will not be fixed with this change as we currently don't have a similar for tenant URLs as well.

Depends on: https://github.com/wso2/carbon-kernel/pull/3358

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -234,7 +234,7 @@
         <cxf-bundle.version>3.3.7</cxf-bundle.version>
         <inbound.auth.oauth.version>6.5.3</inbound.auth.oauth.version>
         <commons-collections.version>3.2.0.wso2v1</commons-collections.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -234,7 +234,7 @@
         <cxf-bundle.version>3.3.7</cxf-bundle.version>
         <inbound.auth.oauth.version>6.5.3</inbound.auth.oauth.version>
         <commons-collections.version>3.2.0.wso2v1</commons-collections.version>
-        <carbon.kernel.version>4.7.0-beta5</carbon.kernel.version>
+        <carbon.kernel.version>4.7.0-beta8</carbon.kernel.version>
```
dewniMW, line 237: need to bump version

AnuradhaSK (approved):

PR #436, opened 2022-07-26
Fix userstore domain incorrect response issue 

shashimalcse:
When user store domain is incorrect, it gives 
```{"schemas":["urn:ietf:params:scim:api:messages:2.0:Error"],"detail":"Cannot add user through scim to user store . SCIM is not enabled for user store <domain name>","status":"500"}```

This issue is fixed in this PR.

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -213,6 +213,10 @@ public class SCIMUserManager implements UserManager {
             user.setUserName(IdentityUtil.addDomainToName(user.getUserName(), userStoreDomainName));
         }
 
+        if (!isUserStoreExist(userStoreDomainName)){
+            throw new BadRequestException("Invalid userstore domain.");
+        }
+
         if (StringUtils.isNotBlank(userStoreDomainName) && !isSCIMEnabled(userStoreDomainName)) {
             throw new CharonException("Cannot add user through scim to user store " + ". SCIM is not " +
                     "enabled for user store " + userStoreDomainName);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java [hunk 1]
```diff
@@ -2721,7 +2721,7 @@ private boolean isMemberAttributeRequired(Map<String, Boolean> requiredAttribute
     }
 
     @Override
-    public void deleteGroup(String groupId) throws NotFoundException, CharonException {
+    public void deleteGroup(String groupId) throws NotFoundException, CharonException, BadRequestException {
```
ShanChathusanda93, line 2724: Changing the public method will lead to some issues with the exsiting usages. The best way I think we can create a separate method with the same impl and add the extra exception.
@senthalan any input on this.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -6081,4 +6084,5 @@ private List<String> addDomainToNames(String userStoreDomainName, List<String> g
         return groupsList.stream().map(groupName -> UserCoreUtil.addDomainToName(groupName, userStoreDomainName))
                 .collect(Collectors.toList());
     }
+
```
ShanChathusanda93, line 6087: Shall we remove this unnecessary extra line.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
senthalan: yes. we can't change the signature of a public method. Can't we throw this as a CharonException with a specific error code and then process in the API layer. 

How we are handling the scim disabled userstores case

shashimalcse:
Will be fixed with https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/437

PR #457, opened 2023-01-30
Fix Client error

VivekVinushanth:
Fixes: 
- https://github.com/wso2-enterprise/asgardeo-product/issues/14403

This PR changes the CharonException which is treated as a server exception. But in this case since the userId is provided by client, which we should consider as client exception (Bad Request with InvaliValue ScimType).

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
@@ -793,7 +793,12 @@ public class SCIMRoleManager implements RoleManager {
                         userStoreManager.getUserListWithID(SCIMConstants.CommonSchemaConstants.ID_URI,
                                 memberObject.get(SCIMConstants.CommonSchemaConstants.VALUE), null);
                 if (isNotEmpty(userListWithID)) {
-                    memberObject.put(SCIMConstants.RoleSchemaConstants.DISPLAY, userListWithID.get(0).getUsername());
+                    String tempDisplay = userListWithID.get(0).getUsername();
+                    if(StringUtils.isNotBlank(userListWithID.get(0).getUserStoreDomain())) {
+                        tempDisplay =
+                                userListWithID.get(0).getUserStoreDomain() + "/" + userListWithID.get(0).getUsername();
+                    }
+                    memberObject.put(SCIMConstants.RoleSchemaConstants.DISPLAY, tempDisplay);
                     memberOperation.setValues(memberObject);
                 }
             }
@@ -816,6 +821,9 @@ public class SCIMRoleManager implements RoleManager {
                 removedMembers.add(memberObject.get(SCIMConstants.RoleSchemaConstants.DISPLAY));
             }
         } catch (UserStoreException e) {
+            if(e.getMessage().equals("Invalid Domain Name")) {
+                throw new BadRequestException("Invalid user store name", ResponseCodeConstants.INVALID_VALUE);
+            }
             throw new CharonException("Error occurred while retrieving the user list for role.");
         }
     }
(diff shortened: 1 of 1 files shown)


GANGANI (approved):

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
```diff
@@ -793,7 +793,12 @@ private void prepareAddedRemovedUserLists(Set<String> addedMembers, Set<String>
                         userStoreManager.getUserListWithID(SCIMConstants.CommonSchemaConstants.ID_URI,
                                 memberObject.get(SCIMConstants.CommonSchemaConstants.VALUE), null);
                 if (isNotEmpty(userListWithID)) {
-                    memberObject.put(SCIMConstants.RoleSchemaConstants.DISPLAY, userListWithID.get(0).getUsername());
+                    String tempDisplay = userListWithID.get(0).getUsername();
+                    if(StringUtils.isNotBlank(userListWithID.get(0).getUserStoreDomain())) {
+                        tempDisplay =
+                                userListWithID.get(0).getUserStoreDomain() + "/" + userListWithID.get(0).getUsername();
```
GANGANI, line 799:
```suggestion
                                userListWithID.get(0).getUserStoreDomain() + "/" + tempDisplay;
```

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
```diff
@@ -857,15 +865,14 @@ private void doUpdateGroups(String roleId, Set<String> newGroupIDList, Set<Strin
...
 
-    private List<String> getUserIDList(List<String> userList, String tenantDomain) throws CharonException {
+    private List<String> getUserIDList(List<String> userList, String tenantDomain) throws BadRequestException {
 
         List<String> userIDList = new ArrayList<>();
         for (String user : userList) {
             try {
                 userIDList.add(getUserIDByName(user, tenantDomain));
             } catch (IdentityRoleManagementException e) {
-                throw new CharonException(String.format("Error occurred while getting the user id " +
```
GANGANI, line 867: Don't we need to handle the server exception that can be happened when trying to resolve ID from userName

VivekVinushanth: The only error thrown from that method implementation is with InvalidRequest. Will add a comment.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
```diff
@@ -816,6 +821,9 @@ private void prepareAddedRemovedUserLists(Set<String> addedMembers, Set<String>
                 removedMembers.add(memberObject.get(SCIMConstants.RoleSchemaConstants.DISPLAY));
             }
         } catch (UserStoreException e) {
+            if(e.getMessage().equals("Invalid Domain Name")) {
```
AnuradhaSK, line 824: Doesn't this exception supports error codes? If supported use error code match.

If the error message has a constant defined in userstore layer, use that constant 

Otherwise update as
`"Invalid Domain Name".equals(e.getMessage())`

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
```diff
@@ -816,6 +821,9 @@ private void prepareAddedRemovedUserLists(Set<String> addedMembers, Set<String>
...
+                throw new BadRequestException("Invalid user store name", ResponseCodeConstants.INVALID_VALUE);
```
AnuradhaSK, line 825:
```suggestion
                throw new BadRequestException("Invalid userstore name", ResponseCodeConstants.INVALID_VALUE);
```

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
VivekVinushanth: Updated.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
VivekVinushanth: It does not have an error code.

AnuradhaSK (approved):

PR #470, opened 2023-04-06
Check users are already avaialble in the group before adding them

sahandilshan:
### Purpose
* Check if the users are already available in the groups before adding them to the group


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -3405,6 +3406,22 @@ public Group patchGroup(String groupId, String currentGroupName, Map<String, Lis
         return getGroup(groupId, requiredAttributes);
     }
 
+    /**
+     * Check whther the users in the set are already existing in the group.
+     *
+     * @param groupId                       Group Id.
+     * @param addedMemberIdsFromUserstore   Set of user ids.
+     */
+    private void removeAlreadyExistingUsersFromTheSet(String groupId, Set<String> addedMemberIdsFromUserstore) throws
```
DMHP, line 3415: Please simplify the name may be as addedMemberIdsFromUserstore > userIdsSet

sahandilshan: Fixed with a2052ac42ef150002e6b6610c01bd529aab4b117

DMHP (approved):

PR #475, opened 2023-07-04
Throw forbidden exception when authorized user doesn't have the required permissions

Yasasr1:
### Purpose 

Only the tenant owner is allowed to remove users from the Administrator role. Framework throws an error if the authorized user doesn't have the permissions when making the request. Previously a 500 error was thrown in this scenario. However 403 should be returned instead of 500.

### Related PRs

- Should be merged first- https://github.com/wso2/charon/pull/385

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
@@ -41,6 +41,7 @@ import org.wso2.carbon.user.core.common.AbstractUserStoreManager;
 import org.wso2.charon3.core.exceptions.BadRequestException;
 import org.wso2.charon3.core.exceptions.CharonException;
 import org.wso2.charon3.core.exceptions.ConflictException;
+import org.wso2.charon3.core.exceptions.ForbiddenException;
 import org.wso2.charon3.core.exceptions.NotFoundException;
 import org.wso2.charon3.core.exceptions.NotImplementedException;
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
```diff
@@ -604,7 +605,7 @@ public Role patchRole(String roleId, Map<String, List<PatchOperation>> patchOper
     }
 
     private void updateUsers(String roleId, String currentRoleName, List<PatchOperation> memberOperations)
-            throws BadRequestException, CharonException {
+            throws BadRequestException, CharonException, ForbiddenException {
```
dewniMW, line 608: changing the method signature of public methods are not practiced

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManager.java
```diff
@@ -545,7 +546,7 @@ public RolesGetResponse listRolesWithPost(SearchRequest searchRequest)
 
     @Override
     public Role patchRole(String roleId, Map<String, List<PatchOperation>> patchOperations)
-            throws BadRequestException, CharonException, ConflictException, NotFoundException {
+            throws BadRequestException, CharonException, ConflictException, NotFoundException, ForbiddenException {
```
dewniMW, line 549: changing the method signature of public methods are not practiced

Yasasr1:
Closing the PR since new commits were not tracked because of a issue with github

PR #481, opened 2023-08-14
Fix for the issue in switching to the latest Java patch version

DilshanSenarath:
This PR includes the fix for the issue in switching to using the latest Java patch versions in the Github Actions.

- https://github.com/wso2/product-is/issues/16437

### DIFF
--- a/.github/workflows/pr-builder.yml
+++ b/.github/workflows/pr-builder.yml
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
-        java-version: [ 11.0.20+8, 17.0.7+7 ]
+        java-version: [ 11, 17 ]
         
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-builder.yml
```diff
@@ -17,9 +17,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    env:
+      JAVA_TOOL_OPTIONS: "-Djdk.util.zip.disableZip64ExtraFieldValidation=true -Djdk.nio.zipfs.allowDotZipEntry=true"
+
     strategy:
       matrix:
-        java-version: [ 11.0.19+7, 17.0.7+7 ]
+        java-version: [ 11.0.20+8, 17.0.7+7 ]
```
chamathns, line 25: we should be able to set the major version here and the action would fetch the latest
```suggestion
        java-version: [ 11., 17 ]
```

chamathns (approved):

PR #494, opened 2023-10-26
Add org perspective SCIM2 servlet mappings

sadilchamishka:
$subject

The organization perspective APIs were referenced by the existing organization allowed APIs
 
Screenshot 2023-10-26 at 06 32 44


AnuradhaSK:
This change will cause the API paths in a way that 

scim2/o/Users etc 
But we want o/scim2/Users
Isn’t it ?

sadilchamishka:
> This change will cause the API paths in a way that
> 
> scim2/o/Users etc But we want o/scim2/Users Isn’t it ?

Sample request
- https://localhost:9443/t/carbon.super/o/scim2/Users

AnuradhaSK (approved):

PR #538, opened 2024-03-13
Rename credential update scenarios

AnjanaSamindraPerera:
## Purpose

$subject

## Related Issue

https://github.com/wso2-enterprise/iam-engineering/issues/963

## Instructions 

Merge after https://github.com/wso2/carbon-identity-framework/pull/5586


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/Scenarios.java
```diff
@@ -26,8 +26,8 @@
  */
 public enum Scenarios {
 
-    CREDENTIAL_UPDATE_BY_ADMIN_VIA_CONSOLE,
-    CREDENTIAL_UPDATE_BY_USER_VIA_MY_ACCOUNT;
+    ADMIN_UPDATED,
+    SELF_UPDATED;
```
malithie, line 31: This is too generic. In future we may also have to know whether if it's a user or an admin updated may be the profile, a particular attribute like email.
And given you are introducing a public constant the room should be available to specify that flow.
Either you can make it an enum with two values which includes the initiated operation and the initiator type, or a one constant which denotes the scenario in a comprehensive way.
Also it's important to capture the fact whether this is a pre or post event (pre credential update by admin or post credential update by admin)

malithie (changes requested):

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/utils/Scenarios.java
AnjanaSamindraPerera: Resolved in [9184f85](https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/538/commits/9184f854cf43604b1242de48b20849916f4e6073)

malithie (approved):

malithie (approved):

PR #590, opened 2025-01-03
Update the group last modified date when updating a group [when unique group id is disabled]

ThaminduDilshan:
## Purpose
When a group is modified by renaming it's display name, adding a user to the group or removing an user from the group, the group last modified date needs to be updated. This PR fixes that issue by updating the group last modified date on such scenarios.

**Note:** The SCIM level update is only required when the group unique id feature is disabled. Otherwise the attributes will be stored in the user store.

## Goals
Update the group modified time when adding/ removing a user to/ from a group.

## Related Issue
- https://github.com/wso2/product-is/issues/19244


ThaminduDilshan:
TODO: Unit tests

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/SQLQueries.java
```diff
@@ -41,10 +41,15 @@ public class SQLQueries {
     public static final String CHECK_EXISTING_ATTRIBUTE_WITH_AUDIENCE_SQL =
             "SELECT TENANT_ID, ROLE_NAME, ATTR_NAME FROM IDN_SCIM_GROUP WHERE IDN_SCIM_GROUP.TENANT_ID=? AND " +
                     "IDN_SCIM_GROUP.ROLE_NAME=? AND IDN_SCIM_GROUP.ATTR_NAME=? AND IDN_SCIM_GROUP.AUDIENCE_REF_ID=?";
+
+    @Deprecated
```
sadilchamishka, line 45: Can't we update this constant. As there is no column called `UM_ATTR_VALUE` which should be anyway wrong.

sadilchamishka (approved):

PR #606, opened 2025-02-05
Fix server error while filtering with special characters

AfraHussaindeen:
### Purpose
- $Subject
- Resolves https://github.com/wso2/product-is/issues/21776

https://github.com/user-attachments/assets/8be4b24f-5c77-4434-bc07-d715228d2ade

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
@@ -608,7 +608,7 @@ public class SCIMRoleManagerV2 implements RoleV2Manager {
             throws BadRequestException {
 
         String searchFilter;
-        String formattedAttributeValue = formatSearchAttributeValue(filterOperation, attributeValue);
+        String formattedAttributeValue = formatSearchAttributeValue(attributeValue);
         
         switch (attributeName) {
             case SCIMConstants.RoleSchemaConstants.DISPLAY_NAME_URI:
@@ -628,13 +628,12 @@ public class SCIMRoleManagerV2 implements RoleV2Manager {
     }
 
     /**
-     * Format the attribute value for search based on the filter operation.
+     * Format the attribute value for search.
      *
-     * @param filterOperation Filter operation
      * @param attributeValue  Value of the attribute
      * @return Formatted attribute value for search
      */
-    private String formatSearchAttributeValue(String filterOperation, String attributeValue) {
+    private String formatSearchAttributeValue(String attributeValue) {
 
         // Handle empty or null values
         if (StringUtils.isBlank(attributeValue)) {
@@ -644,17 +643,7 @@ public class SCIMRoleManagerV2 implements RoleV2Manager {
         // Escape special characters in the attribute value
         String escapedValue = attributeValue.replace("'", "''");
 
-        // Format the search attribute value based on the filter operation
-        switch (filterOperation.toLowerCase()) {
-            case SCIMCommonConstants.CO:
-                return "'%" + escapedValue + "%'";
-            case SCIMCommonConstants.SW:
-                return "'" + escapedValue + "%'";
-            case SCIMCommonConstants.EW:
-                return "'%" + escapedValue + "'";
-            default:
-                return "'" + escapedValue + "'";
-        }
+        return "'" + escapedValue + "'";
     }
 
     private List<RoleV2> getScimRolesList(List<Role> roles, List<String> requiredAttributes)
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
@@ -608,21 +608,53 @@ public class SCIMRoleManagerV2 implements RoleV2Manager {
             throws BadRequestException {
 
         String searchFilter;
+        String formattedAttributeValue = formatSearchAttributeValue(filterOperation, attributeValue);
+        
         switch (attributeName) {
             case SCIMConstants.RoleSchemaConstants.DISPLAY_NAME_URI:
-                searchFilter = ROLE_NAME_FILTER_ATTRIBUTE + " " + filterOperation + " " + attributeValue;
+                searchFilter = ROLE_NAME_FILTER_ATTRIBUTE + " " + filterOperation + " " + formattedAttributeValue;
                 break;
             case SCIMConstants.RoleSchemaConstants.AUDIENCE_VALUE_URI:
-                searchFilter = ROLE_AUDIENCE_ID_FILTER_ATTRIBUTE + " " + filterOperation + " " + attributeValue;
+                searchFilter = ROLE_AUDIENCE_ID_FILTER_ATTRIBUTE + " " + filterOperation + " " + formattedAttributeValue;
                 break;
             case SCIMConstants.RoleSchemaConstants.AUDIENCE_TYPE_URI:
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
```diff
@@ -608,15 +608,17 @@ private String getSearchFilter(String attributeName, String filterOperation, Str
             throws BadRequestException {
 
         String searchFilter;
+        String formattedAttributeValue = formatSearchAttributeValue(filterOperation, attributeValue);
+        
         switch (attributeName) {
             case SCIMConstants.RoleSchemaConstants.DISPLAY_NAME_URI:
-                searchFilter = ROLE_NAME_FILTER_ATTRIBUTE + " " + filterOperation + " " + attributeValue;
+                searchFilter = ROLE_NAME_FILTER_ATTRIBUTE + " " + filterOperation + " " + formattedAttributeValue;
```
PasinduYeshan, line 615: Can you point out where these values are handled later in the flow?
Previously, the attribute value didn't include any operator-specific logic, but now with filteredAttribute, it does. I’d like to understand how this affects subsequent processing. Are there any changes in how these formatted values are utilized in query construction or filtering logic?

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
```diff
@@ -625,6 +627,36 @@ private String getSearchFilter(String attributeName, String filterOperation, Str
...
     }
 
+    /**
+     * Format the attribute value for search based on the filter operation.
+     *
+     * @param filterOperation Filter operation
+     * @param attributeValue  Value of the attribute
+     * @return Formatted attribute value for search
+     */
+    private String formatSearchAttributeValue(String filterOperation, String attributeValue) {
```
sadilchamishka, line 637: Did we test this with multiple filters.

`name co and`

How this is processed through the filtering nodes.
As there is no value after and operation, if we can process `and` part as a value.

AfraHussaindeen: When using a filter like:

`name co and`, 

`and` is treated as part of the search value (not a logical operator).
The function will escape and format `and` as: `'%and%'`. This means the SQL query will match any name column that contains the word `and`. Since no logical operator is present after `and`, it is not treated as part of a condition but rather as a string value inside the search.

AfraHussaindeen:
Upon calling the ` GET /scim2/v2/Roles` API with filtering option, the [listWithGETRole()](https://github.com/wso2/charon/blob/2789631443f1d691c9b1c0ec666b76667f3be855/modules/charon-core/src/main/java/org/wso2/charon3/core/protocol/endpoints/RoleResourceV2Manager.java#L160) method is executed, leading to the creation of a Node associated with the filter ([Node creation](https://github.com/wso2/charon/blob/2789631443f1d691c9b1c0ec666b76667f3be855/modules/charon-core/src/main/java/org/wso2/charon3/core/protocol/endpoints/RoleResourceV2Manager.java#L174)). The [FilterTreeManager](https://github.com/wso2/charon/blob/2789631443f1d691c9b1c0ec666b76667f3be855/modules/charon-core/src/main/java/org/wso2/charon3/core/utils/codeutils/FilterTreeManager.java#L423) class is then used to build the filter tree, where the logic ([Filter Processing](https://github.com/wso2/charon/blob/2789631443f1d691c9b1c0ec666b76667f3be855/modules/charon-core/src/main/java/org/wso2/charon3/core/utils/codeutils/FilterTreeManager.java#L85C9-L124C10)) processes the filter string and generates an intermediate filter expressions list (`tempTokenList`).

For example, if the filter string is "audienceId ne test and displayName co or", the `tempTokenList` will be `["audienceId ne test", "and", "displayName co", "or"]`. Subsequently, another logic step ([Final Token List Generation](https://github.com/wso2/charon/blob/2789631443f1d691c9b1c0ec666b76667f3be855/modules/charon-core/src/main/java/org/wso2/charon3/core/utils/codeutils/FilterTreeManager.java#L140C13-L149C42)) converts this into the final filter expressions list(`tokenList`) as `["audienceId ne test", "and", "displayName co or"]`, ensuring proper expression generation for processing.

Next, the [listRolesWithGET()](https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/blob/253cdeecc799e16f6d4d5f6dcda3cc95fc3d1c7b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java#L350) method is invoked, followed by [buildSearchFilter()](https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/blob/253cdeecc799e16f6d4d5f6dcda3cc95fc3d1c7b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java#L513), which constructs and returns the search filter string (`"audienceId ne test and displayName co or"`) to the **role service layer** ([SCIMRoleManagerV2.java#L498](https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/blob/253cdeecc799e16f6d4d5f6dcda3cc95fc3d1c7b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java#L498)).

Within the **role service layer**, [getExpressionNodes()](https://github.com/wso2/carbon-identity-framework/blob/7443083704a2bc7608d4a9bed51bbd7c11742246/components/role-mgt/org.wso2.carbon.identity.role.v2.mgt.core/src/main/java/org/wso2/carbon/identity/role/v2/mgt/core/RoleManagementServiceImpl.java#L1134) is called, utilizing the [FilterTreeBuilder](https://github.com/wso2/carbon-identity-framework/blob/7443083704a2bc7608d4a9bed51bbd7c11742246/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/FilterTreeBuilder.java#L39) class to construct the filter tree.

The root cause of the internal server error occurs at this stage. Within [FilterTreeBuilder.java](https://github.com/wso2/carbon-identity-framework/blob/7443083704a2bc7608d4a9bed51bbd7c11742246/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/FilterTreeBuilder.java#L89C21-L94C47), the filter expression list(token list) remains unchanged: `["audienceId ne test", "and", "displayName co", "or"]` as no additional processing is defined to handle`and` and`or` correctly when it is ecpected to be value string.
Hence, when the system attempts to build the filter tree, it fails due to unprocessed logical operators (`and`,`or`), leading to an exception.

To resolve this issue, two possible solutions are proposed:

- **Modify the SCIM layer** : Before passing the filter string to the service layer, wrap `or` in single quotes (`'or'`) so that it is correctly treated as an attribute value rather than a operator.
- **Modify FilterTreeBuilder Logic**: Update the `FilterTreeBuilder` class to handle logical operators in a manner similar to `FilterTreeManager`.

The first approach is preferable as it ensures correct filtering logic before reaching the service layer, preventing unnecessary complexities in `FilterTreeBuilder`.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMRoleManagerV2.java
AfraHussaindeen: Refer https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/606#issuecomment-2638754266.

AfraHussaindeen:
Closing this pr as it was decided to proceed with the `Option 2 : Modify FilterTreeBuilder Logic`. For the fix refer https://github.com/wso2/carbon-identity-framework/pull/6440.

PR #609, opened 2025-02-07
Fix user filtering when max user list length exceeds results exist

sadilchamishka:
### Related Issues
- https://github.com/wso2/product-is/issues/22570

When max user list length is set as smaller value, the issue can be easily reproduced.
```
[user_store.properties]
MaxUserNameListLength=3
```

The following flows are fixed in this PR.
1. Filter users without domain. 
2. Filter users with domain

with count param and without count param.

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -1472,6 +1472,9 @@ public class SCIMUserManager implements UserManager {
         domainName = resolveDomainName(domainName, node);
         int totalResults = 0;
         try {
+            if (limit == 0) {
+                limit = getMaxLimit(domainName);
+            }
             // Check which APIs should the filter needs to follow.
             if (isUseLegacyAPIs(limit)) {
                 users = filterUsersUsingLegacyAPIs(node, limit, offset, domainName);
@@ -1548,6 +1551,9 @@ public class SCIMUserManager implements UserManager {
 
     private int getMaxLimitForTotalResults(String domainName) {
 
+        if (StringUtils.isEmpty(domainName)) {
+            domainName = getPrimaryUserStoreDomain();
+        }
         int maxLimit = getMaxLimit(domainName);
         if (isJDBCUSerStore(domainName) && !SCIMCommonUtils.isConsiderMaxLimitForTotalResultEnabled()) {
             maxLimit = Integer.MAX_VALUE;
@@ -1615,11 +1621,7 @@ public class SCIMUserManager implements UserManager {
         String filterOperation = node.getOperation();
         String attributeValue = node.getValue();
 
-        String primaryUSDomain = carbonUM.getRealmConfiguration().getUserStoreProperty(UserCoreConstants.
-                RealmConfig.PROPERTY_DOMAIN_NAME);
-        if (StringUtils.isEmpty(primaryUSDomain)) {
-            primaryUSDomain = UserCoreConstants.PRIMARY_DEFAULT_DOMAIN_NAME;
-        }
+        String primaryUSDomain = getPrimaryUserStoreDomain();
         if ((isFilterUsersAndGroupsOnlyFromPrimaryDomainEnabled()) && !StringUtils
                 .contains(attributeValue, CarbonConstants.DOMAIN_SEPARATOR)) {
             return primaryUSDomain;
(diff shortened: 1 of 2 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -6813,14 +6796,10 @@ private int getMultiAttributeFilteredUsersCount(Node node, int offset, String do
             AbstractUserStoreManager tempCarbonUM = carbonUM;
             // If domain name are not given, then perform filtering on all available user stores.
             while (tempCarbonUM != null && maxLimit > 0) {
-                // If carbonUM is not an instance of Abstract User Store Manger we can't get the domain name.
-                if (tempCarbonUM instanceof AbstractUserStoreManager) {
```
sadilchamishka, line 6817: Removed unnecessary if check of `tempCarbonUM instanceof AbstractUserStoreManager` as it is always an instance of `AbstractUserStoreManager`

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2178,25 +2171,27 @@ private UsersGetResponse getMultiAttributeFilteredUsers(Node node, Map<String, B
...
-            paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder, domainName, maxLimit,
-                    false);
+            int maxLimit = getMaxLimit(domainName);
+            paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder,
+                    domainName, maxLimit, false);
         }
         filteredUserDetails = getFilteredUserDetails(new LinkedHashSet<>(paginatedUserResult.getFilteredUsers()),
                 requiredAttributes);
 
+        int maxLimitForTotalResults = getMaxLimitForTotalResults(domainName);
```
sadilchamishka, line 2181: Use this value of total result calculation logic.
The previous variable is used for the user retrieval process as per line 2174 considering the max user list length of the given user store.

sadilchamishka:
1. Create 5 users which contains "sa"
2. Create a group and assign the users.
3. Add the below property - This limits the number of users listed if count is not specified in the API.
```
[user_store.properties]
MaxUserNameListLength=3
```

Invoke the following queries.
1. Filters users.
https://localhost:9443/scim2/Users?filter=groups+eq+group1+and+username+co+sa

- "totalResults" value is incorrect.

2. Filter users with domain.
https://localhost:9443/scim2/Users?domain=PRIMARY&filter=groups+eq+group1+and+username+co+sa

- "totalResults" value is correct. but "itemsPerPage" is incorrect. It should be 3 as the max user list is 3 when the count is not specified.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1548,6 +1551,9 @@ private boolean isAllConfiguredUserStoresJDBC() {
 
     private int getMaxLimitForTotalResults(String domainName) {
 
+        if (StringUtils.isEmpty(domainName)) {
+            domainName = getPrimaryUserStoreDomain();
+        }
```
sadilchamishka, line 1556: Fix the 1st problem mentioned here https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/609#issuecomment-2644715275

When domain is null, set the primary domain. Then the subsequent logics like `if (isJDBCUSerStore(domainName)` will be evaluated properly.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2178,25 +2171,27 @@ private UsersGetResponse getMultiAttributeFilteredUsers(Node node, Map<String, B
             paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder, domainName, limit,
                     true);
         } else {
-            paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder, domainName, maxLimit,
-                    false);
+            int maxLimit = getMaxLimit(domainName);
+            paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder,
+                    domainName, maxLimit, false);
```
sadilchamishka, line 2176: Fix the 2nd problem mentioned here https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/609#issuecomment-2644715275

When retrieving results when the count is not specified, consider the max limit defined in the corresponding user store domain.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1763,7 +1759,8 @@ private Set<org.wso2.carbon.user.core.common.User> filterUsers(Node node, int of
             throws CharonException, BadRequestException {
 
         // Filter users when filter by group.
-        if (SCIMConstants.UserSchemaConstants.GROUP_URI.equals(((ExpressionNode) node).getAttributeValue())) {
+        if (!SCIMCommonUtils.isGroupBasedUserFilteringImprovementsEnabled() &&
+                SCIMConstants.UserSchemaConstants.GROUP_URI.equals(((ExpressionNode) node).getAttributeValue())) {
```
sadilchamishka, line 1763: Fix the 4th problem mentioned here https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/609#issuecomment-2644715275

We have provided a bypass for filter single attribute of group. It lacks support for pagination and other features (This by pass is used for filter roles from users endpoint, but it is blocked in the master by the given config.)

Hence its better to let groups filter to flow through the below standard filtering logic which works correctly.

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -1472,6 +1472,9 @@ private UsersGetResponse filterUsersBySingleAttribute(ExpressionNode node, Map<S
         domainName = resolveDomainName(domainName, node);
         int totalResults = 0;
         try {
+            if (limit == 0) {
+                limit = getMaxLimit(domainName);
+            }
```
sadilchamishka, line 1477: Fix the 4th problem mentioned here https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/609#issuecomment-2644715275

When the count is not provided, set the max user limit set for the given user store domain.

ashensw (approved):

PR #626, opened 2025-03-26
Remove immutable claims from claims lists for update user in UserManager

osandamaleesha:
#### Description

This PR removes immutable claims from the old claim list and remove the claim difference between old claim list of the user and new claim list of the user

#### Related issue
- https://github.com/wso2/product-is/issues/23644

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -190,6 +190,8 @@ public class SCIMUserManager implements UserManager {
     private static final String SUPPORTED_BY_DEFAULT_PROPERTY = "supportedByDefault";
     private static final String ATTRIBUTE_PROFILES_PROPERTY = "profiles";
     private static final String SHARED_PROFILE_VALUE_RESOLVING_METHOD_PROPERTY = "sharedProfileValueResolvingMethod";
+    private static final String ROLES_CLAIM = "http://wso2.org/claims/roles";
+    private static final String GROUPS_CLAIM = "http://wso2.org/claims/groups";
     private static final String LOCATION_CLAIM = "http://wso2.org/claims/location";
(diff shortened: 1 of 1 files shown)


sadilchamishka (approved):

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -5355,14 +5363,8 @@ private boolean isImmutableClaim(String claim) throws UserStoreException {
...
-                claim.equals(claimMappings.get(SCIMConstants.UserSchemaConstants.USER_NAME_URI)) ||
-                claim.equals(claimMappings.get(SCIMConstants.UserSchemaConstants.ROLES_URI + "."
-                        + SCIMConstants.DEFAULT)) ||
                 claim.equals(claimMappings.get(SCIMConstants.CommonSchemaConstants.CREATED_URI)) ||
-                claim.equals(claimMappings.get(SCIMConstants.CommonSchemaConstants.LAST_MODIFIED_URI)) ||
-                claim.equals(claimMappings.get(SCIMConstants.CommonSchemaConstants.LOCATION_URI)) ||
                 claim.equals(claimMappings.get(SCIMConstants.UserSchemaConstants.FAMILY_NAME_URI)) ||
-                claim.equals(claimMappings.get(SCIMConstants.UserSchemaConstants.GROUP_URI)) ||
                 claim.contains(UserCoreConstants.ClaimTypeURIs.IDENTITY_CLAIM_URI_PREFIX);
     }
```
sadilchamishka, line 5369: This logic is executed before deleting claims. It doesn't allow deleting local claims which are immutable. 

There set of local claims removed from the updating user and existing user. The claim difference are calculated based on that.

From the updating user the required immutable claims are correctly removed, but from the existing user, the immutable claims are not deleted. This PR has fixed that issue. 

Hence the immutable claims will not be detected as a claim difference for update, and the private method not required those claims also. 

The main issue was some immutable claims are detected as claim difference due to above mentioned issue, which was corrected by removing the immutable claims from the old claim list.

PR #635, opened 2025-05-05
Add review suggestions

HasiniSama:
$subject

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
@@ -300,27 +300,12 @@ public class GroupDAO {
     public void updateSCIMGroupAttributes(int tenantId, String roleName,
                                           Map<String, String> attributes) throws IdentitySCIMException {
 
-        doUpdateSCIMGroupAttributes(tenantId, roleName, attributes, SQLQueries.UPDATE_ATTRIBUTES_SQL);
-    }
-
-    /**
-     * Do update SCIM group attributes.
-     *
-     * @param tenantId      Tenant id.
-     * @param roleName      Group name.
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
```diff
@@ -337,7 +322,7 @@ private void doUpdateSCIMGroupAttributes(int tenantId, String roleName, Map<Stri
                 }
                 int[] returnCount = prepStmt.executeBatch();
                 if (log.isDebugEnabled()) {
-                    log.debug("No. of records updated for updating SCIM Group : " + returnCount.length);
+                    log.debug(String.format("No. of records updated for updating SCIM Group: %d", returnCount.length));
```
ashanthamara, line 325: String.format is more expensive than simple concatenation.
Hence let's use string concatenation (+) for simple cases like this as it's faster and lighter.

ashanthamara: We can remove the if condition, as we are doing a simple string concatenation in here where it doesn't affect the perf

HasiniSama: This is now updated.

SujanSanjula96 (approved):

ashanthamara (dismissed):

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/DAO/GroupDAO.java
HasiniSama: Reverting this change as this debug line has a string concat + array length

SujanSanjula96 (approved):

PR #646, opened 2025-05-31
Provide the complex multi valued filtering without any configuration to disable that support

sadilchamishka:
$subject

The identity & non-identity claim filtering performance has been improved recently. Previous implementation cause to delay the response time for number of minutes. As the performance is improved, we don't required to disable the filtering capability.

### Related Issues
- https://github.com/wso2/product-is/issues/24190

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -2304,7 +2304,7 @@ public class SCIMUserManager implements UserManager {
         int totalResults = 0;
         PaginatedUserResponse paginatedUserResult;
 
-        if (limit == 0) {
+        if (limit <= 0) {
           limit = getMaxLimit(domainName);
         }
         paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder, domainName,


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -2306,15 +2304,11 @@ private UsersGetResponse getMultiAttributeFilteredUsers(Node node, Map<String, B
...
 
-        // Handle pagination.
-        if (limit > 0) {
-            paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder, domainName, limit,
-                    true);
-        } else {
-            int maxLimit = getMaxLimit(domainName);
-            paginatedUserResult = getMultiAttributeFilteredUsersWithMaxLimit(node, offset, sortBy, sortOrder,
-                    domainName, maxLimit, false);
+        if (limit == 0) {
```
Malith-19, line 2307: What happen when we receive a negative number. This is not a practical value but in the previous behaviour that scenario was also handled right?

sadilchamishka: Fixed by - https://github.com/wso2-extensions/identity-inbound-provisioning-scim2/pull/646/commits/db21a334c59915ade85461f18edba1a17a00e53a

PR #666, opened 2025-07-02
Refactoring the ask password check

indeewari:
Parent issue

- https://github.com/wso2/product-is/issues/24496


Sachin-Mamoru (approved):

Malith-19 (approved):

amanda-ariyaratne:
What does the `resourceString` looks like?

If this is the payload in the user creation SCIM2 request, then there could be a problematic scenario as follows for migrating users.
In a migrated setup, both SCIM attributes are `urn:scim:wso2:schema:askPassword` and `urn:ietf:params:scim:schemas:extension:enterprise:2.0:User:askPassword` are present. The user creation payload may contain either of these atrributes. According to this logic, the flow will be recognised as an ask password flow only if the payload contains `urn:scim:wso2:schema:askPassword`.

indeewari:
Hi @amanda-ariyaratne ,

We will be addressing this concern + offline invitation with the link in a separate pr.

PR #669, opened 2025-07-03
Removing the account status update claims from the profile update event

indeewari:
Removing the account status update claims from the profile update event

- https://github.com/wso2/product-is/issues/24532

### DIFF
--- a/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
+++ b/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
@@ -5599,7 +5599,7 @@ public class SCIMUserManager implements UserManager {
         carbonUM.setUserClaimValuesWithID(user.getId(), userClaimsToBeModified, null);
 
         // userClaimsToBeModified already includes the userClaimsToBeAdded as well.
-        if (isExecutableUserProfileUpdate && includesOnlyAccountManagementFlowInitClaims(userClaimsToBeModified.keySet())) {
+        if (isExecutableUserProfileUpdate && includesAtLeastOneNonAccountManagementFlowInitClaim(userClaimsToBeModified.keySet())) {
             publishUserProfileUpdateEvent(user, userClaimsToBeAdded, userClaimsToBeModified, claimsDeleted);
         }
     }
@@ -5729,7 +5729,7 @@ public class SCIMUserManager implements UserManager {
 
         // userClaimsToBeModified already includes the userClaimsToBeAdded as well.
         if (isExecutableUserProfileUpdate &&
-                includesOnlyAccountManagementFlowInitClaims(userClaimsToBeModifiedIncludingMultiValueClaims.keySet())) {
+                includesAtLeastOneNonAccountManagementFlowInitClaim(userClaimsToBeModifiedIncludingMultiValueClaims.keySet())) {
             publishUserProfileUpdateEvent(user, userClaimsToBeAdded, userClaimsToBeModifiedIncludingMultiValueClaims,
                     claimsDeleted);
         }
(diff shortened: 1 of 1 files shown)


components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -205,6 +205,9 @@ public class SCIMUserManager implements UserManager {
 
     private static final String MAX_LIMIT_RESOURCE_TYPE_NAME = "response-max-limit-configurations";
     private static final String MAX_LIMIT_RESOURCE_NAME = "user-response-limit";
+    // USER_ACCOUNT_STATE_CLAIM_URIS will be expanded as few other claims to be added.
```
indeewari, line 208: Assuming this list is immutalbe --> add the comment

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -206,6 +206,13 @@ public class SCIMUserManager implements UserManager {
     private static final String MAX_LIMIT_RESOURCE_TYPE_NAME = "response-max-limit-configurations";
     private static final String MAX_LIMIT_RESOURCE_NAME = "user-response-limit";
 
+     /*
+     This will be expanded as few other claims to be added
+     Assuming these claims are immutable and will not be deleted.
+      */
```
malithie, line 212: well this doesn't give the right idea. What you have to say is this list includes claims that can get updated over scim patch or put requests which will be recognized as the certain identity management flow from the IAM product perspective. And when handling such flows using this claim set the assumption is these claims are immutable and they cannot be deleted from a user's profile and only updates are allowed
e.g account enable and disable managed via xxx claim

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -206,6 +206,13 @@ public class SCIMUserManager implements UserManager {
...
+    private static final List<String> USER_ACCOUNT_FLOW_INITIATION_CLAIMS =
```
malithie, line 213: IMO this should indicate it's user account management flow

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
```diff
@@ -7244,4 +7255,30 @@ private String getPrimaryUserStoreDomain() {
...
     }
+
+    /**
+     * Checks whether the given set of claim URIs contains any non-account state claims.
+     * Any non-account state claim is considered as a real profile attribute.
+     *
+     * @param claimUris Set of claim URIs to check.
+     * @return true if there is any non-account state claim, false otherwise.
+     */
+    private boolean hasAnyNonAccountStateClaims(Set<String> claimUris) {
```
malithie, line 7266: This method name should reflect the top comment I added. I think it should be something like includesOnlyAccountManagementFlowInitClaims

components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java
indeewari: Fixed ac2cf820d6ea9f31971e7e71f756ef815fcf8e78

PR #715, opened 2025-08-27
Fix cases where the initiating persona resolves to null when entering a flow

ashanthamara:
In such cases, respective flows are not entered in the Identity context.

## Related Issue
-


ashanthamara:
> Integration Test Runner
> 
> * https://github.com/ashanthamara/product-is/actions/runs/17263452337

image

PR builder passed in the fork

PR #796, opened 2026-06-19
Fix ignoring publishing profile update event for deleting user claims

ashanthamara:
## Purpose
- Previously when the user profile updates only with deleting claims(without modifying claims), the profile update event was not fired. 
- So this PR will fix that issue.

## Related issue
- https://github.com/wso2/product-is/issues/28071


ashanthamara:
Integration Test Runner passed in the local fork:
- https://github.com/ashanthamara/product-is/actions/runs/27807389907

indeewari (approved):

